PR #1, opened 2015-12-15
Mesh animation sample

danke-sra:
Contribution from @fredduarte in SIDIA


danke-sra:
Added a patch to use new loadModel API

liaxim:
The model seems to be too "low" when wearing the headset. Is it me? Other than that, the sample looks and works fine.

thomasflynn:
agree, it does seem low.  it's certainly less scary way down there :)  though we can change it's position in  a future commit.

PR #2, opened 2015-12-16
Modify gvrjassimpmodelloader to load a model from URL

danke-sra:


danke-sra:
Please check URL when testing. The URL does not exist yet -- will work after merging.

danke-sra:
The last commit uses try...catch to handle IOException in onInit. This is because loadModel API throws IOException if the model is not successfully loaded.

Still, the urlBase needs to be changed to the following for testing:
urlBase = "https://raw.githubusercontent.com/danke-sra/GearVRf-Demos/jassimp_network_demo/gvrjassimpmodelloader/assets/";

thomasflynn:
could build and run the sample, but never saw a tree.  got the following log message:

E/JassimpModelLoaderViewManager(23611): Failed to load a model from URL: java.io.IOException: Cannot load a model from path https://raw.githubusercontent.com/gearvrf/GearVRf-Demos/master/gvrjassimpmodelloader/assets/trees/trees9.3ds from NETWORK

ideas?

danke-sra:
Thanks for testing. The URL value in JassimpModelLoaderViewManager.java needs to be manually modified as following (before the patch is merged):

urlBase = "https://raw.githubusercontent.com/danke-sra/GearVRf-Demos/jassimp_network_demo/gvrjassimpmodelloader/assets/";

The URL in the patch is referencing the master branch which does not exist until the patch is merged... Sorry for the confusion.

thomasflynn:
Gotcha, thanks for the correction.  Somehow I misunderstood what was being said earlier.  I've cherry-picked, used the test url, ran and worked fine.  Looks good.

PR #5, opened 2015-12-17
Modify gvrjassimpmodelloader sample to demo skinning from GVRf

danke-sra:


danke-sra:
Added a patch to change to new loadModel API

thomasflynn:
ok, compiles and runs now.

PR #6, opened 2015-12-17
A sample which can contain as many bunnies as we want to make it

dknaan:
A simple sample which can contain as many bunnies as we want to make it complex


liaxim:
It is all good. It does raise the issue of where GVRF is supposed to be relative to the sample's repo but it is outside of the scope of this sample. Everybody will have to update their project reference to GVRF before compiling for now.

Have to figure out how to get write access so I can merge this.

liaxim:
.

PR #11, opened 2016-01-07
Lua scripting sample for GVRf

danke-sra:
- Add GVRScriptManager to manage script engines and
   global variables
- Add library openbeans
- Demo scene creating in scripts
  - Load, transform and add a Collada model
  - Create a text view and add to scene
  - Animate a character


danke-sra:
First version of Lua scripting support. Please comment.

danke-sra:
To get a quick look, jump to the script file is here: https://github.com/danke-sra/GearVRf-Demos/blob/script-dev/gvr-lua/assets/init.lua

The JNLua source tree is used instead of the JAR file because java.beans is not supported by Android (I changed it to openbeans instead).

danke-sra:
Added missing .so file

liaxim:
Looks neat. How do you envision a tighter integration with GVRF?

Like if we have what GearVRLuaScript does embedded in GVRF and, for example, load and execute *.lua from assets? How about some lifecycle methods in the script that GVRF invokes (e.g. onStep)?

Given GVRModelSceneObject, maybe attach a script to it that can easily manipulate the model and respond to events? Etc.

danke-sra:
I think we need to decide whether how much code is stored in each script file: 1) one function like the init.lua, 2) all lifecycle events for a scene object, 3) one script for one scene / application.

It looks like 2) may be the right one. If so, we can add some API like attachScript / detachScript to GVRSceneObject, and make callback to those scripts when onInit / onStep is called.

If scripts are also used to handle other events (touch, eye pointer hit, scene object collision, etc), then, we need to define event interfaces, and find a way to inject script invocations for these interface methods.

Since we may add things like attachScript / detachScript, we may need a framework class GVRScript to hold a script, and may be GVRScriptManager should also load those scripts.

I can collect inputs and do a patch for review.

danke-sra:
Second version uploaded. Some event-delivery mechanism has been added in a framework patch, and this patch depends on that. 

Dependency: https://github.com/Samsung/GearVRf/pull/371

PR #13, opened 2016-01-13
Javascript sample for GVRf

danke-sra:
Demo scene creating in javascript
    - Load, transform and add a Collada model
    - Create a text view and add to scene
    - Animate a character


danke-sra:
The Rhino engine seems to work now. This demo app depends on https://github.com/Samsung/GearVRf/pull/375

PR #14, opened 2016-01-13
adding gvrf widget viewer sample

dattanand:


thomasflynn:
ok, there's about 337 files in assets/data/.  are those needed for this sample?

dattanand:
@thomasflynn not all these are required to run the sample in current form, this is modified version of model viewer, so it carries those assets, I can clean them up as well.

dattanand:
now unnecessary assets have been removed

liaxim:
Ok, the sample needs to be stripped down to just what is needed. There is so much commented out code and code that probably is not needed.

---
All I get in the scene is a stopwatch I can rotate. Is there something special that needs to be done to get to the MyGdxWidget?

dattanand:
there is UI panel on left hand side of watch which can be used to switch between objects, this needs my framework changes which are present in pull request #367 in GVRF.

liaxim:
Ok, so there is a shader compilation error on my S6 and Note5. This is the reason why I didn't see anything. I see the panel on the left but I also see another one on the right (stays empty).

Also I get a crash every time in this scenario: hit the back key (brings up the Oculus prompt), hit back again to return to the app. Could be a gvrf issue but to my knowledge this is the only app exhibiting this behavior.

Let us know what do you intend this sample for - to facilitate the review and presumably get dropped after merging the framework pull request or to be also a genuine sample for end users. If it is a sample for end users it absolutely has to be cleaned up. Thank you.

Now that I can see the panel I can proceed with my review further.

dattanand:
yes the intention of sample is only to verify the framework additions and it can be dropped in current form, I will raise new pull request sometimes later with proper comments and clean code

liaxim:
Thank you for the clarification.

gvrwidgetviewer/project.properties
```diff
@@ -0,0 +1,16 @@
...
+# To customize properties used by the Ant build system edit
+# "ant.properties", and override values to adapt the script to your
+# project structure.
+#
+# To enable ProGuard to shrink and obfuscate your code, uncomment this (available properties: sdk.dir, user.home):
+#proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
+
+# Project target.
+target=android-22
+android.library.reference.1=..\\..\\GearVRf\\GVRf\\Framework
```
liaxim: Please change the path to be unix-style

liaxim:
The sample works fine. The quad could be a little bigger - the colors in the list box are impossible to read. Please squash the commits. Thanks.

thomasflynn:
is this good to go now?

liaxim:
There is a lot of commented out code and ideally the commits should be squashed. Some time ago @dattanand said he plans a cleaner and more documented sample. I'd wait for his feedback on the sample status.

dattanand:
@liaxim I will make the sample clean and squash the commits, I will update it soon

dattanand:
@liaxim I have cleaned up the code and added the new demo. please check if its fine. thanks!!

liaxim:
Code looks good. Merging will have to wait for few days more as we need to convert the Widget plugin also to Android Studio.

dattanand:
@liaxim is there any update on widget plugin android studio migration?

liaxim:
@dattanand Take a look at https://github.com/Samsung/GearVRf/pull/526 and let me know what you think.

dattanand:
@liaxim looks good. eclipse+AS like the way it was done for framework.

dattanand:
does this demo need to be migrated before it gets merged?

liaxim:
We will migrate it after merging. If there are no objections I will merge today after lunch.

dattanand:
yeah fine

PR #15, opened 2016-01-13
Solarsystem beast

ragner:
Just to review


solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
+package org.gearvrf.solarsystem;
```
liaxim: Please use spaces instead of tabs for indentation in this file.

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+
+        private void loadLightMapInfo(GVRContext gvrContext, int lighMapInfoId) {
+            try {
+                InputStream is = gvrContext.getActivity().getResources().openRawResource(lighMapInfoId);
+                int size = is.available();
+                byte[] buffer = new byte[size];
+
+                is.read(buffer);
+
+                is.close();
```
danke-sra: Please move is.close() to finally block (also catch IOException for each close() call).

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+
+        private List<LightMapInfoObject> mMapInfoList;
+
+        public LightMapInfo(GVRContext gvrContext, int lighMapInfoId) {
+            mMapInfoList = new ArrayList<LightMapInfoObject>();
+
+            loadLightMapInfo(gvrContext, lighMapInfoId);
+        }
+
+        private void loadLightMapInfo(GVRContext gvrContext, int lighMapInfoId) {
```
danke-sra: lighMapInfoId is probably a typo "ligh*t*MapInfoId"

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+        public LightMapInfoObject get(int i) {
+            return mMapInfoList.get(i);
+        }
+    }
+
+    private GVRTexture mLightMapTexture;
+    private LightMapInfo mLighMapInfo;
+    private LightMapShader mLightMapShader;
+
+    LightMapManager(GVRContext gvrContext, int lighMapTextureId, int lighMapInfoId) {
```
danke-sra: Please use complete words "ligh*t*MapTextureId" and "ligh*t*MapInfoId"

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+            return mMapInfoList.size();
+        }
+
+        public LightMapInfoObject get(int i) {
+            return mMapInfoList.get(i);
+        }
+    }
+
+    private GVRTexture mLightMapTexture;
+    private LightMapInfo mLighMapInfo;
```
danke-sra: "mLigh*t*MapInfo"?

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+            public String name;
+            public float offsetX;
+            public float offsetY;
+            public float scaleX;
+            public float scaleY;
+        }
+
+        private List<LightMapInfoObject> mMapInfoList;
+
+        public LightMapInfo(GVRContext gvrContext, int lighMapInfoId) {
```
danke-sra: Do you think "lightMapInfoResId" is better? Also, please add "javadoc" to public methods and classes to describe this method and the parameters are used.

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+                lmiObj.offsetX = (float) jsonObj.getDouble("offset.x");
+                lmiObj.offsetY = (float) jsonObj.getDouble("offset.y");
+                lmiObj.scaleX = (float) jsonObj.getDouble("scale.x");
+                lmiObj.scaleY = (float) jsonObj.getDouble("scale.y");
+
+                mMapInfoList.add(lmiObj);
+            }
+        }
+
+        public int lenght() {
```
danke-sra: is "lenght" a typo?

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+    private LightMapShader mLightMapShader;
+
+    LightMapManager(GVRContext gvrContext, int lighMapTextureId, int lighMapInfoId) {
+        loadLightMapTexture(gvrContext, lighMapTextureId);
+        mLighMapInfo = new LightMapInfo(gvrContext, lighMapInfoId);
+        mLightMapShader = new LightMapShader(gvrContext);
+    }
+
+    public void applyLightMap(GVRContext gvrContext, GVRScene scene) {
+        int lenghtLightmapInfo = mLighMapInfo.lenght();
```
danke-sra: is "lenght" a typo?

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+
+        for (int i = 0; i < lenghtLightmapInfo; i++) {
```
danke-sra: It would be better to provide a getter for the `List<LightMapInfoObject>`. Then, the user can use `for (LightMapInfoObject lmiObject : mapInfoList)` to iterate the list. This is more robust and less error-prone.

liaxim: I am missing the point here. LightMapInfoObject is private for a good reason - why would the user know about it, how to use it, what shader type, texture and attributes he or she needs to specify?

danke-sra: I think they need it to get a couple of entries from the loaded file.

solar-system/src/org/gearvrf/solarsystem/LightMapManager.java
```diff
@@ -0,0 +1,181 @@
...
+            sceneObj.getRenderData().getMaterial().setShaderType(mLightMapShader.getShaderId());
+            sceneObj.getRenderData().getMaterial().setTexture(LightMapShader.LIGHTMAP_TEXTURE, mLightMapTexture);
+            sceneObj.getRenderData().getMaterial().setVec2(LightMapShader.LIGHTMAP_OFFSET,
+                    lmiObj.offsetX, lmiObj.offsetY);
+            sceneObj.getRenderData().getMaterial().setVec2(LightMapShader.LIGHTMAP_SCALE,
+                    lmiObj.scaleX, lmiObj.scaleY);
+        }
+    }
+
+    private void loadLightMapTexture(GVRContext gvrContext, int id) {
```
danke-sra: It may be better to use the name resId instead of id.

solar-system/src/org/gearvrf/solarsystem/SolarViewManager.java
```diff
@@ -131,28 +139,31 @@ public void run() {
...
 
         counterClockwise(moonRevolutionObject, 60f);
 
         clockwise(mMainScene.getMainCameraRig().getTransform(), 60f);
 
         counterClockwise(marsRevolutionObject, 1200f);
-        counterClockwise(marsRotationObject, 200f);
+        // counterClockwise(marsRotationObject, 200f);
+
+        new LightMapManager(gvrContext, R.drawable.lightmap_texture, R.raw.lightmap_info).applyLightMap(gvrContext, mMainScene);
```
danke-sra: Does it make sure LightMapManager doesn't get garbage collected? If not, please assign it to a data field and suppress the "unused" warning... Just a suggestion. Sorry about these many comments.

danke-sra:
General comment, the solution looks good. One thing it hasn't demo'ed or proved is the rotation of these scene objects. Does rotation cause a problem for lightmap baking? I will try a bit more with the SDK and give more insightful feedback. Thanks!

danke-sra:
This demo is based on a sample with only moving objects. In this case, the pre-baked lightmap is not always correct. However, it does show how to use lightmap from Beast. I would recommend creating a static scene, and use Beast to create lightmaps for it. It might even include shadows? That would be a nicer demo for Beast.

Other issues are coordinating Beast with dynamic lights. We can review how that is done with other engines and come up with a plan. :)

ragner:
I would like just to show you how the beast integration could work. If it looks ok for you I will work to propose some better and well formated patch :)
Thank you for the review.

ragner:
Today I will start to apply your comments and implement some "final solution".

danke-sra:
Thanks @ragner Look forward to your new work!

solar-system/src/org/gearvrf/solarsystem/SolarViewManager.java
```diff
@@ -181,17 +190,32 @@ private void counterClockwise(GVRSceneObject object, float duration) {
...
-    }
-
     private void clockwise(GVRTransform transform, float duration) {
         setup(new GVRRotationByAxisWithPivotAnimation( //
                 transform, duration, -360.0f, //
                 0.0f, 1.0f, 0.0f, //
                 0.0f, 0.0f, 0.0f));
     }
+
+    public void applySceneLightMap(GVRScene scene, GVRContext gvrContext) {
```
liaxim: Would it make sense for this to be a method of GVRScene or some dedicated lightmap utility in the framework? This method is necessary in every app that uses the lightmap. It could even automatically enable the lightMap for the matching nodes..

liaxim:
Looks good to me.

solar-system/src/org/gearvrf/solarsystem/SolarViewManager.java
ragner: I agree ... looks good to me.
I will implement some logic to do this.

liaxim: Are you planning to do that as part of the existing pull request or maybe later? Thanks.

ragner: If you agree, I would like to do it at a new pull request ...
Thank you.

liaxim: Sounds good.

NolaDonato: There is now a way to integrate lightmaps with dynamic lighting. Multiple light support added in GearVRf 3.0 dynamically creates a shader based on your material and lighting properties. If you have a lightmap texture in your material, it will apply the light map and ignore the light sources.

For applying lightmaps to the objects in the scene, I suggest we make a helper class for doing stuff like that. Maybe it could be a general scene optimization utility. One of the optimizations would be to apply lightmaps instead of using dynamic lighting. It could also do stuff like spatially sort the scene for more optimal picking, combine static geometries, etc.

liaxim:
@ragner Do you have any suggestions what to do about this pull request?

ragner:
Hi @liaxim,
I am going to close it. Thank you.

PR #17, opened 2016-01-15
RenderableView sample to test GVRFrameLayout 

ragner:
This is just to test the result of

https://github.com/Samsung/GearVRf/pull/379


liaxim:
Getting weird effects on my S6. See video: 
[20160118_131702.mp4.zip](https://github.com/gearvrf/GearVRf-Demos/files/95008/20160118_131702.mp4.zip)

If I remove the clocks from the layout it is fine.

liaxim:
@ragner What do you think about the problem I experienced? Would you like me to try something else? I am not sure we can merge the framework pull request until there is some explanation why I got this weird result running the sample.

ragner:
Hi @liaxim, I need check it.
I am on vacation and I will be back at the first week of February.

gvr-renderableview/project.properties
```diff
@@ -0,0 +1,15 @@
...
+# To customize properties used by the Ant build system edit
+# "ant.properties", and override values to adapt the script to your
+# project structure.
+#
+# To enable ProGuard to shrink and obfuscate your code, uncomment this (available properties: sdk.dir, user.home):
+#proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
+
+# Project target.
+target=android-22
+android.library.reference.1=../../GearVRf/GVRf/Framework
```
thomasflynn: please set the library.reference to ../../Framework like the other samples so they compile out of the box without having to modify this.
also, is android-22 necessary or will it work with 19?

thomasflynn:
and yeah, i'm seeing the same thing as @liaxim.  it's as if the viewport or scissor are being modified??

gvr-renderableview/src/org/gearvrf/sample/renderableview/MainActivity.java
```diff
@@ -0,0 +1,63 @@
...
+    private void createView() {
+        mFrameLayout = new GVRFrameLayout(this);
+        mFrameLayout.setLayoutParams(new FrameLayout.LayoutParams(
+        		LayoutParams.WRAP_CONTENT, LayoutParams.WRAP_CONTENT));
+        mFrameLayout.setBackgroundColor(Color.CYAN);
+
+        View.inflate(this, R.layout.activity_main, mFrameLayout);
+    }
+
+    public GVRView getTextView() {
```
thomasflynn: this should probably be renamed to getFrameLayout()?  getTextView seems a little misleading.

thomasflynn:
nice.  thank you.  for your next trick, try passing MotionEvents to widgets inflated inside that FrameLayout so that buttons can be clicked, and lists scrolled.  that would be pretty cool.  it would mean any traditional Android developer could create a 2d ui in VR.

ragner:
Hi @thomasflynn,
I tested the MotionEvents and it works fine at Android 2d coords ... But in my opinion we need some logic to map 2d coods from MotionEvents to 3d coords of VR to get the "touched" GVRViewSceneObject and then pass the MotionEvents to its GVRView.
Does GVRFramework already has something implemented to get the SceneObject touched?

thomasflynn:
You can get the x,y hit location of the picked object in gvrf.  Given that and the width/height of the framelayout, you can compute what the coordinates should be in that layout.  Then you can synthesize click and scroll events by creating your own MotionEvents and passing them to the views.

PR #18, opened 2016-01-22
Spatial Audio example.

thomasflynn:
Spatial Audio example using Google's Cardboard Audio api.

tom.flynn@samsung.com


thomasflynn:
any comments on this?  i'm biased ;-)

gvr-cardboard-audio/src/org/gearvrf/sample/gvrcardboardaudio/SpatialAudioScript.java
```diff
@@ -0,0 +1,164 @@
...
+            // enable lighting for the object
+            cubeObject.getRenderData().setLight(light);
+            cubeObject.getRenderData().enableLight();
+
+            // add the scene object to the scene graph
+            scene.addSceneObject(cubeObject);
+        }
+
+        // add a floor
+		GVRSceneObject floor = new GVRSceneObject(gvrContext, gvrContext.createQuad(120.0f, 120.0f),
```
liaxim: Few lines indented with tabs, if you care about those.

gvr-cardboard-audio/src/org/gearvrf/sample/gvrcardboardaudio/SpatialAudioScript.java
```diff
@@ -0,0 +1,164 @@
...
+    public void onStep() {
+        // Update the 3d audio engine with the most recent head rotation.
+        float headX = cameraRig.getHeadTransform().getRotationX();
+        float headY = cameraRig.getHeadTransform().getRotationY();
+        float headZ = cameraRig.getHeadTransform().getRotationZ();
+        float headW = cameraRig.getHeadTransform().getRotationW();
+        audioEngine.setHeadRotation(headX, headY, headZ, headW);
+
+        // rotate the cube
+        cubeObject.getTransform().rotateByAxis(0.5f, 1.0f, 0.0f, 0.0f);
```
liaxim: Needs null check. The cubeObject is not created if the robot loads.

liaxim:
There is a bug apparently in some Eclipse ADT plugin (like mine). It would not let me install from eclipse because of the .so in audio.jar. I have the option "Force error ..." off of course. The command line build works fine. We can wait until/if someone complains and then move the .so out of the jar like I did or document this.

The fallback cube - is it needed? It shows half-buried into the "ground" for me - not that it matters so much but fyi.

--
Other than that it works nicely. I'd say the null check is the only thing of importance.

thomasflynn:
And this is why we have code reviews :).  Nice feedback.  The fallback cube isn't really needed.  I first developed this with that cube.  When when Danke fixed the model loading so i could load the model properly, i switched to the model.  Since loading the model apparently requires a try-catch, i put the cube in there.  I can just remove it.

As for the jar file, yeah, i build from the command line.  I'm finding things seem to build faster there than in Eclipse.  It's taken from the cardboard sdk, so I guess it works fine with Android Studio?  If so, I think maybe we can document the issue for now.  We need to move to Studio before SDC.  It's going to be disruptive and annoying, but we need to make the move.

rahul27:
I tried it this morning and got the same error on eclipse:

[2016-02-04 09:41:51 - SpatialAudioActivity] The library 'audio.jar' contains native libraries that will not run on the device.
[2016-02-04 09:41:51 - SpatialAudioActivity] Additionally some of those libraries will interfer with the installation of the application because of their location in lib/
[2016-02-04 09:41:51 - SpatialAudioActivity] lib/ is reserved for NDK libraries.
[2016-02-04 09:41:51 - SpatialAudioActivity] The following libraries were found:
[2016-02-04 09:41:51 - SpatialAudioActivity]  - lib/armeabi-v7a/libvraudio_engine.so

How do you run command line?

thomasflynn:
same way you do any android project.  first generate the build.xml:

android update project --path `pwd`

once that's done, you don't need to do it again.  then from then on, you can just do:

ant build

gvr-cardboard-audio/project.properties
```diff
@@ -0,0 +1,15 @@
...
+# To customize properties used by the Ant build system edit
+# "ant.properties", and override values to adapt the script to your
+# project structure.
+#
+# To enable ProGuard to shrink and obfuscate your code, uncomment this (available properties: sdk.dir, user.home):
+#proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
+
+# Project target.
+target=android-19
+android.library.reference.1=../../Framework
```
rahul27: You might want to preempt https://github.com/gearvrf/GearVRf-Demos/pull/22 and fix the path beforehand.

danke-sra:
I reviewed the code. The audio position integration between Cardboard SDK and GVRf is very good.

thomasflynn:
removed the cubeObject and the tabs.  I could import into Eclipse, but couldn't build.  For some reason it was complaining about all the GVRf apis.  Saying that GVRActivity and other classes didn't exist.  Like it didn't know where the library was.  But it clearly does.  Too tired to fight with Eclipse now.  Maybe tomorrow morning.

liaxim:
Builds from eclipse (except the known issue with deployment because of the .so in the jar). Runs nicely.

PR #22, opened 2016-01-24
Fix android project properties for samples.

johnspurlock:
Update the android library reference in 39 of the 40 sample projects,
reflecting the fact that the sample repo is now more likely to be in
a peer directory to the main repo, rather than a subdirectory.

GearVRf-DCO-1.0-Signed-off-by: John Spurlock
john.spurlock@gmail.com

### DIFF
--- a/collision/project.properties
+++ b/collision/project.properties
@@ -12,4 +12,4 @@
 
 # Project target.
 target=android-19
-android.library.reference.1=../../Framework
+android.library.reference.1=../../GearVRf/GVRf/Framework
--- a/eye-picking/project.properties
+++ b/eye-picking/project.properties
@@ -12,4 +12,4 @@
 
 # Project target.
 target=android-19
-android.library.reference.1=../../Framework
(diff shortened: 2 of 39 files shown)


thomasflynn:
hm, that's something to discuss.  i've been checking out the demos repo to a directory peer to the old Samples/ directory which is at the same level as Framework/.  Perhaps we should go this direction.  will discuss with the team.

johnspurlock:
Usually it's simpler to avoid nested git repos.

For what it's worth, one of the samples already assumes this new structure [1].  The PR here simply makes the same change for the other samples.

Hope that helps,
- john

[1] https://github.com/gearvrf/GearVRf-Demos/blob/master/complexScene/project.properties

thomasflynn:
alright, thanks.

thomasflynn:
talked with the team.  they agreed this would be more straight-forward for users.  it's getting late and i'm getting tired.  i'll probably merge this tomorrow.

ragner:
:+1:

PR #25, opened 2016-02-05
Add a nicer icon.

thomasflynn:
Was previously using the default Android icon.  Now it will have a gvrf
icon.

tom.flynn@samsung.com


liaxim:
Does it really have to be 2000 × 1818? Do we care?

thomasflynn:
reduced the icon size.

PR #28, opened 2016-02-06
Adding remote scripting sample.

thomasflynn:
Adding remote scripting sample.  It enables the debug server and sets up
a text object with the ipaddress of the phone so we know where to telnet
into.

tom.flynn@samsung.com


danke-sra:
Looks good. It should process IP address reliably, and it's good that the GVRTextViewSceneObject can be referenced.

PR #41, opened 2016-02-29
Update javascript sample for script bundle

danke-sra:
- Added a gaze cursor
- Use a script bundle to handle GVRScript events and ISensorEvents
  for a scene object.
- The text changes color when clicked


danke-sra:
Updated JS demo to show how to use multiple scripts. Need this patch to compile: https://github.com/Samsung/GearVRf/pull/435

gvr-javascript/assets/script.js
```diff
@@ -1,29 +1,41 @@
...
+  mainScene.addSceneObject(boat);
 
   // Text
-  var textView = utils.newTextViewSceneObject(gvrf, "GVRf scripting in Javascript");
+  var textView = new GVRTextViewSceneObject(gvrf, gvrf.getActivity());
+  textView.setText("GVRf scripting in Javascript");
+  textView.setRefreshFrequency(GVRTextViewSceneObject.IntervalFrequency.HIGH);
   var textSize = textView.getTextSize();
+  textView.setName("text");
   textView.setTextSize(textSize);
```
liaxim: Please set the text size to something in the 6-8sp range. Looks decent. textView.getTextSize() returns number of pixels but textView.setTextSize() works with scaled pixels and thus setTextSize(getTextSize()) blows up the text. Also please use var textView = new GVRTextViewSceneObject(gvrf) instead of var textView = new GVRTextViewSceneObject(gvrf, gvrf.getActivity()).

danke-sra:
Updated JS to reflect GVRTextViewSceneObject changes. Will do an API update for GearVRLua after that patch is merged.

thomasflynn:
tested.  works.  though, it looks like it might be slow (i see several (4-5) cursors when moving my head).  is this sample slow? or has gvrf become slow somehow?

gvr-javascript/assets/script.js
```diff
@@ -1,30 +1,43 @@
...
-  if (animations.length != 0) {
-      animations[0].setRepeatMode(1).setRepeatCount(-1);
-      animations[0].start(animationEngine);
+  var animations = boat.getAnimations();
+  if (animations.size() != 0) {
+      animations.get(0).setRepeatMode(1).setRepeatCount(-1);
+      animations.get(0).start(animationEngine);
   }
+
+  mainScene.setFrustumCulling(false);
```
liaxim: Is there a particular reason why this is necessary?

danke-sra: Thanks for asking. Animation will change the bounding box and culling currently doesn't work with that.

gvr-javascript/src/org/gearvrf/sample/gvrjavascript/GearVRJavascriptScript.java
```diff
@@ -16,18 +16,81 @@
...
+                GVRSceneObject cursor = new GVRSphereSceneObject(context);
+                GVRRenderData cursorRenderData = cursor.getRenderData();
+                GVRMaterial material = cursorRenderData.getMaterial();
+                material.setShaderType(shaderManager.getShaderId());
+                material.setVec4(CustomShaderManager.COLOR_KEY, 1.0f, 0.0f,
+                        0.0f, 0.5f);
+                mainScene.addSceneObject(cursor);
+                cursor.getRenderData().setDepthTest(false);
+                cursor.getRenderData().setRenderingOrder(100000);
+                controller.setSceneObject(cursor);
```
rahul27: Since the cursor is slow and laggy. You can get rid of the cursor by removing the controller.setSceneObject(cursor); call . You will still get the events for textview without the need for a visual cursor.

rahul27:
I noticed the lag as well. We could see if it is noticeable for an object added to the camera rig.

As Mihail said, the lag is considerably reduced when the animations are disabled. For now we could remove the visual cursor since we really only need key events rather than a visual gaze cursor.

thomasflynn:
i'd like to understand why animations slow things down so much.  something is obviously bogging down the GL thread.  what is it? and can it be moved off the GL thread?  (even if we don't do that moving for this check-in)

danke-sra:
I will wait for some definitive suggestion. I think if cursor is too slow, can I use a default texture for cursor, instead of a shader?

rahul27:
Yeah. You could try a texture. The current cursor also has an alpha value that may be unnecessary in your case.

thomasflynn:
i want to stress that what concerns me is the performance and whatever is causing the performance slow down, not the fact the cursor doubles/triples.  If performance was fixed, the cursor would be fine.  I want to make sure we identify the underlying cause of why the performance is low.  Visually, it is a simple enough scene that it shouldn't be a problem.

danke-sra:
This change and corresponding framework change didn't add anything to the GL thread. The performance thing might come from animation and rendering as usual. I didn't experience lags on my side. Let me double check.

danke-sra:
Update on the lag. It is because this scene is complex. It is one of the most complex animated scene I can find. I tested Rahul events sample with the same framework changes, and there is no performance difference than before. It uses the same cursor. So the only thing is the scene. I will test the other animation sample and see if there is a difference.

thomasflynn:
yeah, i wasn't implying that this change introduced a performance degradation, merely exposing one that is probably inherent in the framework (which concerned me).  in any case, both this and the associated framework change need rebase.

PR #42, opened 2016-03-03
Updates triggered by the GVRTextViewSceneObject improvements

liaxim:
This depends on https://github.com/Samsung/GearVRf/pull/440

### DIFF
--- a/gvr-javascript/assets/script.js
+++ b/gvr-javascript/assets/script.js
@@ -11,8 +11,7 @@ function onInit(gvrf) {
 
   // Text
   var textView = utils.newTextViewSceneObject(gvrf, "GVRf scripting in Javascript");
-  var textSize = textView.getTextSize();
-  textView.setTextSize(textSize);
+  textView.setTextSize(6);
   textView.getTransform().setPosition(0, 0, -2);
 
   mainScene.addSceneObject(textView);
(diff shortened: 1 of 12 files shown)


thomasflynn:
sorry.  i didn't see this comment and merged the other commit.  can you go ahead and update the text size in another commit?  i need to merge this one now.

danke-sra:
It seems the unit changed, so the size is still OK. If the new unit makes more sense, we can stick to it. However, it is a risky change if unknown apps are using it....

liaxim:
I am not sure which unit you are referring to. The textSize has always been in scaled pixels.

Yes, it is not a perfect change but the previous version of the textView was more or less broken.

liaxim:
The values were eye balled before, eye balled now. Instead of the gigantic 2000x1000 default canvas&buffer size we have more appropriate canvas that also respects the aspect ratio of the quad. Hence the difference in font size.

danke-sra:
So the original text size is pixel in the internal buffer. This was a bad design. Internals should be transparent for text size too. If there is not a big change in look of apps, it should be fine. Have we documented how to select the text size, and is it unchanged for later internal changes too?

liaxim:
The text size is in scaled pixels but that text was drawn onto a 2000x1000 canvas previously hence it required large values to produce decently sized text.

Selecting the text size is not documented, it was eye balled and continues to be. I believe with the recent change it becomes a bit more predictable and consistent.

danke-sra:
Anyway, it is an improvement on the text view. Good work.

PR #57, opened 2016-03-31
gvr-javascript: show handling IActivityEvents with a script

danke-sra:


danke-sra:
This patch shows how to handle GVRActivity events and how to print from a script. It's quite simple. I am not sure if it is useful, but just submit it.

PR #59, opened 2016-04-02
Adding some simple javascripts as examples

thomasflynn:
Adding some simple javascripts as examples.  The photo and video samples were originally developed by @danke-sra, i only had minor modifications in them.

tom.flynn@samsung.com


danke-sra:
Looks good to me. Will let others review it more.

PR #62, opened 2016-04-09
Multilight

NolaDonato:
Multiple lighting sample program


liaxim:
Looks like DirectLight.java is not needed. It probably should be removed so as not to confuse people.

liaxim:
Looks good to me and works fine.

Got confused a bit because I didn't realize how far the backdrop was away from the model so I didn't quite get the position of the spots. We can address minor things like missing braces later. Nothing that would prevent it from merging.

liaxim:
Can we remove beach.zip, sphere.obj? The jni folder probably can be cleaned up too. Again, something we can do afterwards.

liaxim:
Monoscopic works fine with multiple lights too.

NolaDonato:
I pushed a fix removing these files.

PR #67, opened 2016-04-20
Solar system sample converted to Android Studio

liaxim:


liaxim:
I find it beneficial to have gradle-2.2.1 installed somewhere and configure AS to use it. The version of the android plugin for gradle that we use requires it.

PR #68, opened 2016-04-20
Scene objects sample converted to Android Studio

liaxim:


liaxim:
@thomasflynn Tried opening this one with AS 2.0. Prompted me for gradle - specified 2.2.1. Then asked me to upgrade because the android plugin used is too old - told it to remind me later. Rebuilt fine and ran. We might have to look closer at your environment.

PR #70, opened 2016-04-20
Model-viewer sample converted to Android Studio

liaxim:


liaxim:
Just fyi: I am following Jason's template and replacing the files with these from the sample I am modifying. This template defines two modules: the sample and the framework. The framework is built as a dependency for the sample. Importing and converting the sample's Eclipse project in AS will not just work.

PR #90, opened 2016-04-25
Rename gvr-multilight-androidstudio to gvr-multilight

liaxim:
+ remove local.properties and few other generated files


liaxim:
Planning to merge this after lunch, in case someone might have feedback.

liaxim:
Ok, this is not necessary apparently. Somehow the old dir was still showing up for me but it was my local workspace. After everything was synced it went away.

PR #78, opened 2016-04-21
converted applications to android studio

NolaDonato:
Converted the following applications to android studio build (the originals are removed)
collision
complexScene
eye-picking
gamepad 
gvrbulletsample
gvrcamera2renderscript
gvrcockpit_v1_5_0
gvr-cubemap             
gvrjassimpmodelloader   
gvr-litshader           
gvr-multilight          
gvr-meshanimationsample 
simple-gallery          
simple-sample

PR #91, opened 2016-04-25
converted applications to android studio; rebased

liaxim:
These are Nola's commits from her pull request: https://github.com/gearvrf/GearVRf-Demos/pull/78

Reverted some of the previous commits I made, will sanity check this first, squash the commits and merge.


liaxim:
Actually squashing is not going to be that good.

PR #99, opened 2016-04-26
added cusor and edittext.

thomasflynn:
added cusor and edittext.  the accuracy on picking the 'update' button
in the edittext isn't very good, but will work if you focus on the 'up'
portion of the button.  This should bring the repository up to date with
what I will present at SDC.

Also, DisplayUtils and the CustomPostEffectShaderManager looked to be in
their original directory.  was that deliberate?  or were they forgotten?
I moved them just in case.

tom.flynn@samsung.com


liaxim:
The files must have been overlooked.

PR #108, opened 2016-04-28
Reduce immersivepedia's apk size to 65MB

liaxim:


liaxim:
Splash screen enabled

PR #109, opened 2016-05-02
Fixed the events sample

rahul27:
* Replaced sphere object with a quad
* Removed the custom shader
* Fixed picking on the Android View

r.rudradevan@samsung.com


liaxim:
Looks good to me. Will merge tomorrow after lunch if there are no objections.

thomasflynn:
verified.  looks good to me.

PR #119, opened 2016-05-04
Gvrcubemap build fixed

liaxim:
- added copies of the missing shaders from gvrf; Nola will have to update with the right version
- added on touch handling so you can just cycle over all available environments


rahul27:
:thumbsup: Looks good.

liaxim:
Thanks @rahul27 for the verification. I will merge this in the afternoon.

liaxim:
@NolaDonato Please check out the shaders after you come back.

PR #126, opened 2016-05-06
Convert gvrwidgetviewer to Android Studio

liaxim:


gvr-widgetviewer/app/build.gradle
```diff
@@ -0,0 +1,29 @@
+apply plugin: 'com.android.application'
+
```
jason2kim: jsut be consistent with the module name, would it be better?
gvr-widgetviewer/app/build.gradle => gvr-widgetviewer/gvrwidgetviewer/build.gradle 

The same with settings.gradle

liaxim:
Addressed Jason's concern about the module name.

jason2kim:
it looks good to me

PR #131, opened 2016-05-09
gvrjassimp: package name correction

liaxim:
Not sure this is the only problem with the sample. I was getting weird compilation errors implying the libraries from the libs folder were not picked up but then the errors suddenly stopped and I can't get them to reappear no matter what.


thomasflynn:
yep, builds and runs for me now.

PR #146, opened 2016-05-12
Made gvr-bullet sample build by adding libbullet.so in jniLibs

NolaDonato:
The sample now builds with Android Studio. The .so files need to be under jniLibs. Still doesn't build under Eclipse for me.


NolaDonato:
It builds under Eclipse but does not find the bullet library and crashes on startup. Eclipse is supposed to have a feature that lets you specify external libraries but these menu items do not appear when using the Android tool chain :-(

PR #149, opened 2016-05-13
reverted to single light version

NolaDonato:
Reverted back to the version which uses C++ shaders instead of Java shaders


thomasflynn:
@vittal does this look good to you?

PR #150, opened 2016-05-16
Restored cubemap demo to single light

NolaDonato:
Restore the single light version of cubemap demo to be compatible with
previous versions of GearVRf


gvr-cubemap/gvrcubemap/src/main/java/org/gearvrf/cubemap/CubemapActivity.java
```diff
@@ -27,15 +25,6 @@
...
-    }
-
-    @Override
-    public boolean onTouchEvent(MotionEvent event) {
-        if (event.getAction() == MotionEvent.ACTION_UP) {
-            script.onTouch();
-            return true;
-        }
-        return super.onTouchEvent(event);
+        setScript(script, "gvr_note4.xml");
```
thomasflynn: shouldn't this be gvr.xml?

PR #153, opened 2016-05-17
Added libbullet.so in another place

NolaDonato:
Some phones want it in jniLibs/armeabi and some want it in jniLibs/armeabi-v7a


thomasflynn:
yay, it works now :)

PR #156, opened 2016-05-17
gvr-events: fixed sample to process all motion and key events

rahul27:
Also bumped the gradle version

r.rudradevan@samsung.com


thomasflynn:
looks good to me.  sample runs well.  any other comments from anyone?

PR #162, opened 2016-05-24
gvr-performance: Modified the sample to use ZipLoader

rahul27:
r.rudradevan@samsung.com


NolaDonato:
Thanks Rahul for updating the sample! I tested it with your zipload patch to GearVRf and everything looks good.

PR #164, opened 2016-05-25
App module name fixes

jason2kim:
Unique application name .apk will be picked from Jenkins script build

GearVRf-DCO-1.0-Signed-off-by: Jason Kim
jason2.kim@samsung.co


thomasflynn:
needs rebase.

jason2kim:
Will do soon. It requires additional verification from someone as it changes many files.

liaxim:
Alternative way to fix the apk names is by adding this to the module's gradle file:

```
    applicationVariants.all {variant ->
        variant.outputs.each { output ->
            output.outputFile = new File(
                    output.outputFile.parent,
                    rootProject.name + ".apk")
        }
    }
```

The benefit of having this generically named "app" module is that it provides a predictable directory layout. Any automated operations would benefit by not having to discover the sample's module name.

jason2kim:
It seems to be a good solution to me as well

jason2kim:
the new patch: https://github.com/gearvrf/GearVRf-Demos/pull/194

PR #176, opened 2016-06-01
Use GVRVideoSceneObject for GVR-Video app

deepakrawat22:
Changed GVR-Video app to use GVRVideoSceneObject, earlier it used to create SurfaceTexture by itself and called updateTexImage() on non-GL Thread

deepak.rawat@samsung.com


NolaDonato:
Thanks for cleaning up the demo, it looks a bit nicer. I tested it on the Note4 and it works with pull request #595 now whereas the previous version did not.

PR #181, opened 2016-06-03
Fix gvr-jassimp timing error

NolaDonato:
The sample starts all the animations i the call to getNextMainScene. If the asset loader returns with the model right away this will work. If not, the animations will be started before they are loaded.

I moved the animation startup to after the model loading. This guarantees that all the animations are loaded. All the textures may not be loaded yet.

Eventually we should change this sample to use GVRAssetLoader.loadModel and pass the GVRScene so it will wait for all the textures.


liaxim:
There are changes to gvr-polyline and gvr-3dcursor. Are they necessary?

NolaDonato:
The changes to gvr-3dcursor are not really changes - Git merge just deleted the entire contents of the file and replaced it with the same thing. I am not sure how I got it to do that.
I have removed the gvr-polyline .project file. I have not done a pull request for gvr-polyline yet.

liaxim:
If you'd like you could checkout the previous version of the gvr-3dcursor files so you won't have them changed in this pull request:

```
git checkout <commit-id> <path-to-file>
```

Since this really changes a single file you might want to consider squashing the commits.

liaxim:
Depends on https://github.com/Samsung/GearVRf/pull/597

liaxim:
Works fine with https://github.com/Samsung/GearVRf/pull/597. Can be merged after it is.

PR #185, opened 2016-06-06
Immersivepedia - move some of the processing off the gl thread

liaxim:
- should help with really bad stuttering when passing over the totems

- avoid ConcurrentModificationException when iterating over interactiveObjects
- preload the loadingComponent's mesh and textures instead of loading them on every pass over the totem
- move to a background thread some of the audio-related processing

--
This is just a first pass. There might be subsequent changes in other PRs.

---


liaxim:
Regardless of framework changes, I still think this might be a useful change as it will increase the drawing fps.

PR #187, opened 2016-06-07
gvr-complexscene: change for the cases where adding a shader

liaxim:
must happen on the gl thread


liaxim:
Closing for now since deadlock has been encountered.

PR #191, opened 2016-06-09
gvr-360video: set the apk name by including a pregenerated build.xml

liaxim:


liaxim:
If this works out will make the change everywhere.

liaxim:
This is for the case where we have to use the old build toolchain.

PR #194, opened 2016-06-10
Module name change and unique apk naming

parthmehta209:
1. Changed the module name of all samples to be app
2. Added a script in each app/build.gradle file for apk names to be generated using the directory  name of the sample.
3. Removed .iml files that were committed


parthmehta209:
@jason2kim @liaxim please review.

jason2kim:
Oh, I see, you reverted the existing modules to ‘app’.

parthmehta209:
@jason2kim Yes. I changed all the sample apps to have a module called app. To make all the apps uniform.

jason2kim:
That should be okay. The existing Eclipse users just need to delete and re-import the apps because of the new directory path.

jason2kim:
The apk name is correctly generated from AS build, for example gvr-sample.apk
However, the unaligned apk still has ‘app’, like app-debug-unaligned.apk. We can avoid copying this in Jenkins batch build if we want.

jason2kim:
Did some verifications myself. It looks good to me.

liaxim:
Will merge in the afternoon.

parthmehta209:
@liaxim Moved all build.xml files added in https://github.com/gearvrf/GearVRf-Demos/pull/193 from <modulename>/src/main/ to app/src/main.

Issue #46, opened 2016-03-10
gvr-Vuforia teapot slide issue.


PsycoTodd:
Hello,

Thank you for the nice demo! However, I notice that in the gvr-Vuforia demo, when the user hold the image target and slightly shift the HMD, the virtual teapot shifts more which makes it not relatively fixed at the detected position. If the user move the card away from the camera, the teapot disappear.

For the disappearance issue, It is possible that the teapot is moved behind the quad of video, or just out of the Gear VR culling Frustum. In the original Vuforia VR demo, when drawing the background video quad, depth test is disabled so no matter how the object which is drawn after the video quad moves in z direction, it will not be hide. So my question is: how can we increase the Z-Far plane for larger culling Frustum, and how can we disable the depth test for the screen object to make sure that the 3D things drawn are all above the video?

And for the 3D object slightly shift during moving the image target or move the head, is that possible due to the view port difference? 

Hope to hear from you soon!

liaxim:
See GVRSceneObject.getRenderData().setDepthTest and GVRScene.getMainCameraRig().setFarClippingDistance.

PsycoTodd:
Hello,

I notice that the translation and scale of the virtual model is the major issue of the current demo.

Scaling: I notice that in the code an empirical UV coordinate is used to draw the background video, this UV has effect of zoom in. However, the original scale obtained from Vuforia does not take this into account so it may lead to the virtual object looks smaller relative to the background video.

Translation: aka the sliding issue. the object does not move to the center as the image. If I draw a cube at the center of the Vuforia tracking target, when I move the image target and watch the virtual object drawn in Gear VR, the virtual object won't keep at the center of the image target at all. It is like that the video is projected by the vuforia matrix and then projected by the Gear VR system, while in the demo, the virtual teapot is manipulated by vuforia model view but project only by gear VR matrix.
Not quiet sure the explanation about this issue but hope some one can share their opinion on this.

Todd

PsycoTodd:
OK, I think the the root of evil is the empirical UV coordinate system, it makes the translation and scaling factor of the vuforiaMVMatrix not aligned with the rendered result. Depending on the UV sub sample region, the vuforiaMVMatrix need to be updated accordingly.

Hope it helps!

liaxim:
@PsycoTodd We will be looking at this but since the problem is with the sample itself it may take a while. Free free to open a pull request if you have a fix in mind.

PsycoTodd:
[...]

PR #201, opened 2016-06-16
Fix for teapot slide issue #46

deepakrawat22:
This pull request fixes the teapot  slide issue. Following changes made this possible:

1 - Updated the sample files from vuforia to latest.
2 - Custom shader for AR objects as they use MVP matrix obtained from vuforia task.
3 - Corrected the camera pass through object dimensions. The size ratio of GL viewport and camera passthrough must match.

deepak.rawat@samsung.com


liaxim:
What is the exact fix for the slide?

--
I think having the camera passthrough as a full screen quad yields a very distorted image while wearing the headset. I still notice sliding going on but it is harder to tell. What was wrong with the old passthrough dimensions? The teapot does not move smoothly anymore, it is kind of stuttering.

deepakrawat22:
The fix for the slide is to pass the MVP matrix obtained from vuforia, so that's why the custom shader. Another thing that's need to be done is to make the GL viewport match the one obtained from vuforia which depends on camera passthrough.

liaxim:
@deepakrawat22 Do you think we should try it out and review again?

deepakrawat22:
@liaxim I am still doing some changes, will commit them in some time and let you know once its ready for review. Thanks.

deepakrawat22:
@liaxim this pull request is ready for review now. Need the PR #630 from GearVRf to test the new commit. Thanks.

gvr-vuforia/app/src/main/java/org/gearvrf/vuforiasample/VuforiaSampleMain.java
```diff
@@ -169,7 +187,12 @@ private void hideTeapot() {
         }
     }
 
-    public void updateObjectPose(State state) {
+    public void updateObjectPose() {
+        State state = mRenderer.begin();
+
+        // viewport as per default 1024 x 1024 window size
+        mainScene.getMainCameraRig().setViewport(0, 224, 1024, 576);
```
liaxim: How did you determine these numbers?

liaxim:
This is much better.

@teresa-mao If you have time please check this out.

@thomasflynn You might want to see how this works now.

gvr-vuforia/app/src/main/java/org/gearvrf/vuforiasample/VuforiaSampleMain.java
deepakrawat22: The camera resolution is 2560 X 1440 giving aspect ratio as 16:9. The GL viewport aspect ratio also match. So just change 1024 X 1024 to 16:9

liaxim:
Teresa noticed that the aspect ratio of the camera view is really off in monoscopic mode.

deepakrawat22:
Finally corrected the code and now generate texture co-ordinates based on device aspect ration. Now its working on both S7 and Note4. Still need GearVRf #630 to work.

deepakrawat22:
This pull request is not yet ready for review as need to change the setting the viewport technique.

deepakrawat22:
@liaxim Removed the viewport set call as viewport is set from config file.

liaxim:
@teresa-mao Please let us know how does it work for you. Thanks.

liaxim:
Looks good. The teapot could move a bit smoother. On Note5 the teapot is flipping oddly occasionally; not observed on S7. We can continue improving as the time permits.

liaxim:
Will merge in the afternoon if there are no objections.

teresa-mao:
yeah we should merge it & continue to make improvements later.

PR #207, opened 2016-06-22
Stopping the app process when back is pressed. This cleans up issues …

wkudaimi:
…with audio continuing to play and dinaosarus disappearing on app relaunch.


liaxim:
Will merge lunch time

liaxim:
There is a concern since we are supposed to show the Oculus prompts on back key. Will try something else and update.

liaxim:
@wkudaimi I created https://github.com/gearvrf/GearVRf-Demos/pull/215 which is a bit nicer about the whole killing process. I think your change would work too but since I verified 215 as working in the various cases where the Oculus menus may or may not show I'll close this pull request and merge 215.

Issue #210, opened 2016-06-23
gvr-video: Radiosity Effect stops working if change to IMAX mode and back


deepakrawat22:
I observed that Radiosity effect won't work if switch to IMAX mode in theater environment 2 and come back to normal mode.

deepakrawat22:
#212 fixed this issue

PR #212, opened 2016-06-24
Fix for radiosity after switch to IMAX (#210)

deepakrawat22:
The mTransitionWeight  was large and radiosity shader was ignoring due to condition.

deepak.rawat@samsung.com


liaxim:
Looks good to me. Will merge after lunch.

PR #259, opened 2016-07-18
Demo apps uses framework and extenion aar files

deepakrawat22:
GearVRf-DCO-1.0-Signed-off-by: 
Deepak Rawat deepak.rawat@samsung.com
nola.donato@samsung.com Nola Donato
Abhijit Jagdale


liaxim:
Let's squash the commits, please. Let me know if you need assistance doing that. Thank you.

liaxim:
I am not so sure about the expected framework-debug.aar name but it will do for now. Let's have the commits squashed and if @NolaDonato is ok with this we will merge.

gearvrf-libs/ReadMe.txt
```diff
@@ -0,0 +1 @@
+Copy the framework and extentions (if sample app uses extention) aar files here.
```
rahul27: should be *extension*

gvr-360photo/app/build.gradle
```diff
@@ -21,16 +30,17 @@ android {
...
+            output.outputFile = new File(
+                    output.outputFile.parent,
+                     rootProject.name + ".apk")
+        }
+    }
 }
 
 dependencies {
-    compile project(':framework')
+    compile(name:'framework-debug', ext:'aar')
```
rahul27: Shouldn't we be using the latest release?

https://github.com/Samsung/GearVRf/releases/tag/v3.0.1

rahul27:
This change would make it difficult to test framework changes with the sample apps. I would have to keep replacing the aar file every time I make a change. The alternative would be replace the framework dependency for all apps I would want to test.

I think it should be possible to create different build variants for the sample apps. Gradle allows you to do that. The default could be set to **release** which should use the aar file. Developers could switch to the **debug** build variant that defaults to "../../GearVRf/GVRf/Framework/framework".  Here are a few links that might help:

https://developer.android.com/studio/build/build-variants.html
http://stackoverflow.com/questions/22439593/different-dependencies-for-debug-and-release-in-gradle-and-android-studio

gvr-360photo/app/build.gradle
liaxim: It should be a generic name until we have a way to pick all aars from a location.

liaxim:
Fair point but the first concern is to make the samples as easy as possible for end users, as soon as possible.

The build variants are neat but if there is an explicit framework module defined that points to the framework's tree then people will have to deal with setting up that regardless of the build variants. Or go in there and manually remove the module..

For testing changes this actually makes it easier since you can drop your aar in the dir and have any sample pick it up. The build times are cut significantly too.

Making changes to the framework and debugging is a different story and we will come up with a better solution in due time.

rahul27:
This change still needs you to create the "framework-debug.aar" which will requires the framework to be set up and build correctly. Unless there is a way to easy download it and place it in the libs dir?

rahul27:
The readme will also need updating in light of this change:

https://github.com/gearvrf/GearVRf-Demos/blob/master/README.md

liaxim:
Yes. People are supposed to get the latest release from github: https://github.com/Samsung/GearVRf/releases/tag/v3.0.1. The whole idea is for people not to have to deal with building the framework.

We will have a nightly build release on github too.

rahul27:
Just renaming the gvrf-3.0.1.aar to framework-debug.aar won't work. You still need to add the oculus files:

- The build.gradle would need to to be updated to support the jar(SystemUtils.jar and VrApi.jar) files:
````
dependencies {
    compile fileTree(dir: 'libs', include: ['*.jar'])
    compile(name:'framework-debug', ext:'aar')
}
```
- You will also need to create a jniLibs folder and place the libvrapi.so file there.

Only then can would the app build.

liaxim:
The Oculus-related manual work is inevitable. We cannot bundle those binaries in. We have it in the instructions.

PR #272, opened 2016-07-25
gvr-tutorial-lesson6: remove an image using invalid characters;

liaxim:
can't build otherwise


liaxim:
@NolaDonato Removed кк.jpg from under the assets folder. "aapt" doesn't like cyrillic characters. Doesn't seem to be used by anything.

PR #273, opened 2016-07-27
valueFloatTexture must be set when create circle alpha

deepakrawat22:
If not set first time render result in error and error shader is called

GearVRf-DCO-1.0-Signed-off-by: 
Deepak Rawat deepak.rawat@samsung.com


liaxim:
Looks good to me. Will merge after lunch.

PR #284, opened 2016-08-01
Minor gradle improvements

liaxim:
- introduce a common gradle file that is included in all samples
- add a common task to copy the oculus signature files


liaxim:
If there are no comments will merge on Wednesday

PR #288, opened 2016-08-03
Fixed camera position in x3d-demo and testmodels

NolaDonato:
If the camera is too close to the object, the x3d-demo will reposition it to view the entire model. Also sets the near and far clipping planes from the x3d file.

testmodels scaling was fixed so objects are closer to the camera


liaxim:
@NolaDonato There are no changes in this pull requests (Files changed 0).

NolaDonato:
Oops. There are files there now.

liaxim:
My understanding is that Mitch was going to look at the problem this tries to solve. For now not planning to merge this unless I hear updates.

mwitchwilliams:
This code was causing issues.  It was overriding the location of a camera specified in the X3D scene.
We will have some new code likely by September's end that will give X3D more control over the camera and switching camera positions (part of X3D's <Anchor> tag).
Designers may also want a 3d mesh where the camera exists if for example, they want to switch from camera position to camera position, having a 3d mesh marking each camera's location.

liaxim:
Ok then. If there are no objections I will close the pull request tomorrow.

PR #292, opened 2016-08-05
Move some of the common build properties to common.gradle

liaxim:
- some of the changes can be now made to a single file and that
will reflect across all the samples


liaxim:
If there are no comments will merge Thursday morning

PR #293, opened 2016-08-05
gvr-camera2renderscript: add model SM-N9208 as one needing

liaxim:
mNeedYuvConversion too


liaxim:
If there are no comments will merge after lunch.

PR #294, opened 2016-08-06
Fix gvr-javascript text animation problem

NolaDonato:
Fixes problem cause by PR #776
 The script bundle is attached really early in onCreate and it can't find
 the "text" scene object when it tries to attach the bundle. I changed
 one of the scripts to attach "text.js" to the "text" scene object
 programatically instead of in the bundle. Now the onSensorEvent and
 onStep functions get called.


liaxim:
Will merge after lunch if there are no objections. Will deal with the onSensorEvent separately.

PR #295, opened 2016-08-07
upgrade all samples to 2xMSAA

thomasflynn:
Most of our samples had MSAA turned off.  We can get 2xMSAA for free.
Enable that.

tom.flynn@samsung.com


liaxim:
If there are no comments will merge after lunch.

PR #296, opened 2016-08-08
Modified the X3D demo to test the sensors

rahul27:
r.rudradevan@samsung.com


rahul27:
I've fixed the demo. Both this and https://github.com/Samsung/GearVRf/issues/789 are good to go.

liaxim:
What is supposed to happen here? If I go over the box it starts getting larger and larger until it finally disappears for good. Is that it?

Mitch, are you ok with this change? Also, please consider adding a way to easily switch between the various x3d files in the demo. Maybe use volume up and down to go through the list of files.

rahul27:
Yup. That is the behavior defined in the x3d test file.

liaxim:
Ok, looks good to me then. Mitch, please let us know if you are ok with merging this.

mwitchwilliams:
Yes, just reviewed the code and it looks good to me.
That box moves from z=-10 to z = 60, and indeed ends up out of view behind the camera, but its movement leaves the impression it is getting larger.  The camera sits at  (0, 0, 0).  Should be able to see the box if the user looks in the opposite direction.

liaxim:
Ok, will merge after lunch.

PR #298, opened 2016-08-15
Fix use of picking

NolaDonato:
The samples were attaching the picker to the wrong object which worked
before because the picker always assumed the camera. Now the picker
generates the ray from whereever it is attached and defaults to the
camera if it has no owner.


liaxim:
Looks good. The affected samples work as expected. Will merge after lunch if there are no further comments.

PR #299, opened 2016-08-16
add the framework outputs directory

thomasflynn:
I probably missed the memo due to travel and other chaos i was handling,
  but would this commit be useful?  I'm adding the relative path to
  where the framework aar should be given the way we usually clone the
  repos.

tom.flynn@samsung.com

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -2,7 +2,7 @@ allprojects {
     repositories {
         jcenter()
         flatDir {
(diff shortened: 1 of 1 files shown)


liaxim:
There is the LIBS_DIRECTORY property that if configured would copy the aar to gearvrf-libs. But this additional thing is useful too. Will merge after lunch if there are no comments.

PR #300, opened 2016-08-17
fix shader for gvr-outline sample

roshanch:
It doesn't like gl_FragColor.

GearVRf-DCO-1.0-Signed-off-by: Roshan Chaudhari roshan.c@samsung.com

### DIFF
--- a/gvr-outline/app/src/main/java/org/gearvrf/gvroutline/OutlineShader.java
+++ b/gvr-outline/app/src/main/java/org/gearvrf/gvroutline/OutlineShader.java
@@ -37,10 +37,10 @@ public class OutlineShader extends org.gearvrf.GVRShaderTemplate
 
     private static final String FRAGMENT_SHADER =
               "precision mediump float;\n"
-            + "out vec4 gl_FragColor;\n"
+            + "out vec4 outColor;\n"
             + "uniform vec4  u_color;\n"
             + "void main() {\n"
-            + "  gl_FragColor = u_color;\n"
+            + "  outColor = u_color;\n"
             + "}\n";
 
     public OutlineShader(GVRContext gvrcontext)


gvr-outline/app/src/main/java/org/gearvrf/gvroutline/OutlineShader.java
```diff
@@ -37,10 +37,10 @@
 
     private static final String FRAGMENT_SHADER =
               "precision mediump float;\n"
-            + "out vec4 gl_FragColor;\n"
+            + "out vec4 outColor;\n"
```
liaxim: For completeness should the shader also specify version 300 explicitly?

roshanch: We are adding that in shaderTemplate.java

liaxim:
Will merge in the afternoon if there are no further comments.

thomasflynn:
since the PR went up after the build team enabled the pre-merge build polling, i'm assuming we should a pass/fail from the build system sometime soon?  (since it polls every hour)?

liaxim:
Ok, will wait for the build result.

thomasflynn:
just talked to Shreeti.  it doesn't work the way one would hope.  because IT won't allow proper communication with github (*frustration scream*), what she set up will only trigger a build _after_ we click 'merge pull request'.  Then, we'll know _after the fact_ if the build got broken.  It's better than nothing, but is still close to nothing :(.

liaxim:
Texture is missing. Will investigate tomorrow.

roshanch:
There was issue in batching code, we set material to null in renderCamera() but batching code was setting it to renderpass(0) material, and we don't update material from render-pass if its not null in non-batch mode. PR #775 is updated.

liaxim:
Verified. Will merge after lunch.

PR #304, opened 2016-08-23
Build tweaks for modelviewer2

liaxim:
including the widgetplugin aar for now since it is not part
of the automated build; will work with the build team on
fixing that


liaxim:
Merging early as this is the only way to test it

PR #306, opened 2016-08-24
FIxed null pointer crash in gvr-meshanimation

NolaDonato:
This construct does not work. The animation engine pointer is null here. I am not sure what changed to make it not work now but it seems suspicious to me. I changed the sample to start the animation engine as soon as it is constructed. Not sure if this is the right fix but it avoids the crash.

GVRScene mainScene = gvrContext.getNextMainScene(new Runnable() {
            @Override
            public void run() {
                mAssimpAnimation.start(mAnimationEngine);
            }
        });


liaxim:
Needs rebase. Also the commits do not have any gvr-meshanimation changes but a bunch of modelviewer2 ones.

There is something funky going on in the viewManagers that we have. This sample works just fine with the pull request for the backend/framework separation - https://github.com/Samsung/GearVRf/pull/832.

liaxim:
@NolaDonato What should we do about this pull request? Thanks.

NolaDonato:
You said it worked OK for you on the latest master so I think we should close it.

PR #308, opened 2016-08-25
Add a project that builds all samples

liaxim:


liaxim:
Since the projects are independent gradle's parallel module builds are enabled; speeds up the process nicely.

``./gradlew assembleDebug`` would build all the samples we have
``./gradlew installDebug`` would install all the samples to the connected phone

I'd imagine you may want to remove all the samples from the phone easily too; there is no gradle task for this unfortunately but while in the GearVRf-Demos directory you can run this:
``grep -r package --include="AndroidManifest.xml"|grep -oE '"org.gearvrf.*'|xargs -n 1 adb uninstall``
This should remove all but the accessibility sample.

liaxim:
Using this you can have all the samples built and installed on a phone in 5 minutes (using Linux workstation of course). So whenever necessary going through all the samples for verification is quite painless now.

liaxim:
Will merge tomorrow morning if there are no comments.

PR #309, opened 2016-08-25
Looks like gvr-keyboard depends on the order of execution of

liaxim:
onStep and drawframeListeners on startup; we have not defined that
so you can't assume anything


liaxim:
Will merge after lunch if there are no comments

PR #312, opened 2016-08-29
gvr-cardboard-audio: update to use the audio engine from googlevr

liaxim:


liaxim:
This is simple enough change. If there are no comments will merge in the afternoon.

PR #315, opened 2016-09-06
Using GVRMain in place of GVRScene to be consistent

mwitchwilliams:
with other GVR Demos.  Also removed references to
GVRAnimation since that is handled in X3Dobject.

GearVRf-DCO-1.0-Signed-off-by: Mitch Williams 
m1.williams@partner.samsung.com

### DIFF
--- a/gvr-x3d-demo/app/src/main/java/org/gearvrf/x3ddemo/X3DparserScript.java
+++ b/gvr-x3d-demo/app/src/main/java/org/gearvrf/x3ddemo/X3DparserScript.java
@@ -19,44 +19,29 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
 
 import org.gearvrf.FutureWrapper;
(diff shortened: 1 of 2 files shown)


gvr-x3d-demo/app/src/main/java/org/gearvrf/x3ddemo/X3DparserScript.java
```diff
@@ -70,21 +55,13 @@
  *
  */
 
-public class X3DparserScript extends GVRScript
+public class X3DparserScript extends GVRMain
```
rahul27: You might want to rename this call to X3DparserMain since it is no longer a script.

You could do it in a future pull request.

rahul27:
Looks good. Tested it and it works fine.

gvr-x3d-demo/app/src/main/java/org/gearvrf/x3ddemo/X3DparserScript.java
mwitchwilliams: Yes, I was think that made sense to rename it:
X3DParserMain.java to be consistent with other demo’s.

Glad you mentioned it.

PR #318, opened 2016-09-09
gvr-360video: expand the sample to support playback via the exoplayer

liaxim:
Also the playback is paused and resumed when the activity is paused/resumed.


liaxim:
Depends on https://github.com/Samsung/GearVRf/pull/845

liaxim:
Will merge after lunch if there are no objections.

PR #321, opened 2016-09-13
Modified Samples for IO integration

rahul27:
r.rudradevan@samsung.com

### DIFF
--- a/gvr-3dcursor-simple/app/src/main/java/org/gearvrf/io/cursorsimple/CursorMain.java
+++ b/gvr-3dcursor-simple/app/src/main/java/org/gearvrf/io/cursorsimple/CursorMain.java
@@ -20,7 +20,6 @@ import android.graphics.Color;
 
 import org.gearvrf.GVRAndroidResource;
 import org.gearvrf.GVRBitmapTexture;
-import org.gearvrf.GVRComponent;
 import org.gearvrf.GVRContext;
 import org.gearvrf.GVRMain;
 import org.gearvrf.GVRMaterial;
@@ -33,6 +32,7 @@ import org.gearvrf.GVRTexture;
 import org.gearvrf.GVRTransform;
 import org.gearvrf.io.cursor3d.Cursor;
 import org.gearvrf.io.cursor3d.CursorManager;
+import org.gearvrf.io.cursor3d.IoDevice;
 import org.gearvrf.io.cursor3d.MovableBehavior;
 import org.gearvrf.io.cursor3d.SelectableBehavior;
 import org.gearvrf.io.cursor3d.SelectableBehavior.ObjectState;
@@ -42,7 +42,8 @@ import org.gearvrf.scene_objects.GVRModelSceneObject;
 import org.gearvrf.utility.Log;
 
 import java.io.IOException;
-import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This sample can be used with a Laser Cursor as well as an Object Cursor. By default the Object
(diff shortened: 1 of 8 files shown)


rahul27:
@parthmehta209

gvr-3dcursor/app/build.gradle
```diff
@@ -27,6 +27,7 @@ if(file("../../../../GearVRf-Demos/common.gradle").exists()) {
 android {
     defaultConfig {
         applicationId "org.gearvrf.io.cursor"
+        //applicationId = "com.samsung.android.gvr.sample"
```
parthmehta209: Is this needed ?, if yes should we put _VENDOR_TODO_ before it.

gvr-3dcursor/app/src/main/assets/settings.xml
```diff
@@ -226,11 +227,6 @@
             position="center"
             themeId="crystal_sphere"
             type="object">
-        <!--_VENDOR_TODO_ add your first device to the settings.xml here-->
```
parthmehta209: Why not just add the left_INDEX here, and add right_INDEX at line 256. Do we need to enable two other cursors for testing the IODevices. This will also need a note at 226 to change active=no to active=yes. Also names Left Cursor and Right Cursor are repeated which might be confusing in the settings UI.

gvr-3dcursor-simple/app/src/main/java/org/gearvrf/io/cursorsimple/CursorMain.java
```diff
@@ -63,10 +64,25 @@ public void onInit(GVRContext gvrContext) {
...
+        /*
+        TemplateDevice device1 = new TemplateDevice(gvrContext, "template_1", "Right controller");
+        TemplateDevice device2 = new TemplateDevice(gvrContext, "template_2", "Left controller");
+        devices.add(device1);
+        devices.add(device2);
+        */
+
+        /*
+        HandTemplateDevice device = new HandTemplateDevice(gvrContext, mainScene);
+        devices.addAll(device.getDeviceList());
```
parthmehta209: Can add HandTemplateDevice device, and device1, and device2 as member variables. So trying out the template device will only involve un-commenting code. Right now when you uncomment device.close(). It does not recognize device as its not a member variable.

parthmehta209:
Tested both samples with template hands, works well.

gvr-3dcursor/app/build.gradle
rahul27: It's not needed. Will get rid of it.

gvr-3dcursor-simple/app/src/main/java/org/gearvrf/io/cursorsimple/CursorMain.java
rahul27: I can do that but it just increases the number of steps required for integration.

gvr-3dcursor/app/src/main/assets/settings.xml
rahul27: Simpler to uncomment this block and test the cursors than to play around with the existing settings. I've added the note regarding line 226 to the integration steps.

I've also changed the cursor names.

gvr-3dcursor-simple/app/src/main/java/org/gearvrf/io/cursorsimple/CursorMain.java
parthmehta209: its fine as it is then.

gvr-3dcursor/app/src/main/assets/settings.xml
parthmehta209: ok

parthmehta209:
This should be good to go.

PR #325, opened 2016-09-19
Add command line option to enable building the demos with an

liaxim:
alternative backend; by default oculus will be used


liaxim:
To build with an alternative backend use ``./gradlew -Pno_oculus=true assembleDebug``

liaxim:
Will merge tomorrow if there are no comments.

PR #336, opened 2016-09-28
gvr-VRGallery: tweaks so it builds

liaxim:
seems to run ok


thomasflynn:
looks good

PR #337, opened 2016-09-28
Modified all shader code for multitex:

dnvasilev:
vec4 a_tex_coord => vec2 a_texcoord


liaxim:
Works as expected. Will merge together with the fw pull request.

PR #338, opened 2016-10-03
Dynamically generate the list of demo projects

liaxim:


liaxim:
Will merge tomorrow morning if there are no objections.

thomasflynn:
This is cool :)  thank you :+1:

liaxim:
Unrelated but: ``./gradlew uninstallDebug`` removes all apps from the phone.

There is also ``uninstallAll`` for all build variants but we usually don't need this.

To get launcher refresh and remove some of the leftover app icons you may need to:
- reboot the phone
- go to AppManager, find TouchWiz Home, force stop

PR #339, opened 2016-10-03
Switching samples from GVRScript to GVRMain

liaxim:
Depends on https://github.com/Samsung/GearVRf/pull/876

- gvr-opacityanigallery, gvr-sample - changed to restore what I believe is the intended behavior
- gvr-simplegallery - the tiles change size after a bit but it is down to the app; still debating whether this app shouldn't be removed altogether since gvr-opacityanigallery is its superset.

GearVRf-DCO-1.0-Signed-off-by: Mihail Marinov m.marinov@samsung.com


liaxim:
gvr-opacityanigallery: further clarification; Based on the code it is supposed to show one scene object out of a ring buffer of N. On both the currently and previously selected scene objects opacity and scale animations are applied (this is the only difference from gvr-simplegallery). Cycling through the list happens by looking up and down or looking down and up.

liaxim:
gvr-accessibility shows a not yet known problem. Have to dissect its onInit as it is doing lots of things.

liaxim:
What is left are the libgdx samples. The plugin mechanism the libgdx extension uses is rather fragile.

liaxim:
Ready for review

PR #341, opened 2016-10-06
Updated the google vr libraries to v1

rahul27:
r.rudradevan@samsung.com


liaxim:
Please check gvr-cardboard-audio. There are multiple definitions of classes due to audio classes being in the daydream aar. I guess we need to switch to whatever comes with daydream but
- some code changes are needed for the sample
- we need a solution when we build with the other backends but not with daydream

PR #344, opened 2016-10-08
Update gvr-shadows demo

NolaDonato:
Use a moving spotlight instead of a moving directional light.
Change light orientation so shadows are more obvious.


liaxim:
There is a change to gvr-modelviewer2 that looks totally unnecessary. If there is no good reason for it please back it out.

NolaDonato:
I am still reworking the shadows sample. Closing this PR, will reopen a new one later,.

PR #345, opened 2016-10-12
Using WearTouchPad in GearVRf-Demos

parthmehta209:
depends on https://github.com/Samsung/GearVRf/pull/889


thomasflynn:
ok, good, i can confirm this is functional now.  +1, looks good to me.

liaxim:
Will merge Tuesday after lunch unless there are comments.

PR #350, opened 2016-10-20
gvr-solarsystem: fix the sun position

liaxim:
This is Roshan's fix from couple of months back.

Previously the sun was centered at the origin (the camera was inside the sphere) so that lead to:
- no sun visible because the back faces are culled
- a draw call no matter where you would look at


thomasflynn:
yep, this fixes it.  why did it get changed in the first place?

liaxim:
I do not know but will find out.

liaxim:
@thomasflynn The proper fix plus explanation why it got changed: https://github.com/gearvrf/GearVRf-Demos/pull/352

PR #352, opened 2016-10-24
gvr-solarsystem: correct fix for the sun not being visible

liaxim:

### DIFF
--- a/gvr-solarsystem/app/src/main/java/org/gearvrf/solarsystem/SolarMain.java
+++ b/gvr-solarsystem/app/src/main/java/org/gearvrf/solarsystem/SolarMain.java
@@ -15,22 +15,25 @@
 
 package org.gearvrf.solarsystem;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-
 import org.gearvrf.GVRAndroidResource;
+import org.gearvrf.GVRCamera;
+import org.gearvrf.GVRCameraRig;
 import org.gearvrf.GVRContext;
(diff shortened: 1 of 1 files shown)


liaxim:
The bug was triggered by the https://github.com/Samsung/GearVRf/commit/0d984e7d7eb677082dcdd0393a85399cbee0da65 framework change. Since the parent was now being checked, the moonRevolutionObject scene object was causing an exception because it was (implicitly) trying to add the mainCameraRig owner scene object to a different parent (it gets the root scene object as parent automatically).

The quick fix for the sample was to remove completely the moonRevolutionObject and it can easily be mistaken as superfluous. But the camera rig (point of view) is supposed to be positioned at the moon.

These changes restore the sample to what I believe is its intended behavior. Creating a new camera rig is clunky but feels the correct solution instead of obscure code that removes the existing camera rig from its parent and implicitly have it assigned a new parent.

I already have a note that creating a proper camera rig can be problematic. Also clunky. addChildObject(Component c) is rather bad api I will target for deprecation.

liaxim:
For reference this is where the moonRevolutionObject got lost: https://github.com/gearvrf/GearVRf-Demos/commit/aabfc0284a20a30411a8bab7fc39eac213d2b460. Can be used to see the original version on which this fix is based off.

rahul27:
@liaxim :smile: 

I removed it since it was not explicitly added to the scene.

liaxim:
Will merge tomorrow morning if there are no objections.

PR #360, opened 2016-11-08
3DCursor sample: Small change to the cursor activation listener

rahul27:
r.rudradevan@samsung.com


parthmehta209:
Tested this change. This can be merged.

PR #371, opened 2016-11-16
Include 64bit binaries only when the custom ARM64 property is defined

liaxim:


liaxim:
@dnvasilev This should fix your issue once it gets merged in.

liaxim:
Will merge in the afternoon

PR #373, opened 2016-11-21
Modified the mesh animation sample

rahul27:
Added a few more animations
Also added an animation that makes use of the Skeletal
Framework

Use tap to change the animation.

r.rudradevan@samsung.com


rahul27:
Depends on https://github.com/Samsung/GearVRf/pull/904

rahul27:
Will reopen later.

PR #382, opened 2016-12-02
Use binaries from docs.gearvrf.org

parthmehta209:
GearVRf-DCO-1.0-Signed-off-by: Parth Mehta <parth.mehta@samsung.com>


liaxim:
We need an option to choose maven or local aars. Default to maven.

liaxim:
Please make similar changes to the build files under GearVRf-Demos/old/.

liaxim:
Looks good. Testing..

PR #385, opened 2016-12-08
fix NPE in gvr-3dcursor

parthmehta209:


gvr-3dcursor/app/src/main/java/org/gearvrf/io/cursor/CursorMain.java
```diff
@@ -570,9 +570,11 @@ public void onActivated(Cursor newCursor) {
         @Override
         public void onEvent(CursorEvent event) {
             GVRSceneObject sceneObject = event.getObject();
-            SpaceObject spaceObject = objects.get(sceneObject.getName());
-            if(spaceObject != null) {
-                spaceObject.handleCursorEvent(event);
+            if(sceneObject != null) {
```
liaxim: Why would an event has no object associated with it?

liaxim:
Had a chat with @parthmehta209. It is ok for now.

PR #390, opened 2016-12-16
Use the framework's default gvr.xml where appropriate

liaxim:


liaxim:
Rebased

PR #392, opened 2017-01-03
Adding the daydream sample

rahul27:
r.rudradevan@samsung.com


thomasflynn:
@NolaDonato, thoughts on the SampleMain.java in this example?  Can it be simplified with newer APIs that we have?  or is this good as is?

NolaDonato:
I tested this on my old Motorola Droid Ultra and it worked! Good job. Please change the call to loadFutureTexture to use the asset loader: 
            sphere = new GVRSceneObject(gvrContext, gvrContext.loadFutureMesh(new
                    GVRAndroidResource(gvrContext, "sphere.obj")), gvrContext
                    .getAssetLoader().loadFutureTexture(new GVRAndroidResource(gvrContext, "texture.png")));

I am trying to stamp out usages of the context loading routines in favor of those same routines on the asset loader because I will be removing them from the context eventually.

rahul27:
@NolaDonato I've changed the calls to use assetLoader.loadFutureTexture.

liaxim:
Having a backend-specific demo is problematic in my opinion. Some of the specific reasons:

What happens when the demo is run on a phone on which we use some of the other backends first?
Should we expose ways for apps to tweak the backend priorities?
Should we build this sample with the daydream backend exclusively?

---
I propose extending the existing eye picking sample so that it supports the controller and the extended functionality if it happens to be available or otherwise fall back to the default gaze-based behavior.

rahul27:
Here are the answers to your questions:

_What happens when the demo is run on a phone on which we use some of the other backends first?_
*   It should run fine without errors. You would not be able to use it since it requires the controller.

_Should we expose ways for apps to tweak the backend priorities?_
*   We could, but I created this app just to set a template to start developing daydream applications on GVRf. It automatically picks up daydream on daydream ready phones. We could add this feature in the future if needed.

_Should we build this sample with the daydream backend exclusively?_
*   We could, but it is not necessary at the moment.

_I propose extending the existing eye picking sample so that it supports the controller and the extended functionality if it happens to be available or otherwise fall back to the default gaze-based behavior._

That is fine, but the problems I see is:

If an app developer wants a sample that describes only picking he/she would use the eye picking sample but the sample also includes code for the daydream controller that he/she may have to strip of if not needed. It also adds the extra layer of complexity required to understand the code that does both gaze and controller based picking therefore making the sample much more than what it was actually created for.

Alternatively I could add GearVR based gaze functionality into this app but that again may be undesirable for developers looking only to build daydream applications.

thomasflynn:
In the short term, we do need a sample that shows people how to use the DayDream controller with GearVRf.  Ideally, in the future, the code for that will be abstracted out in the IO Framework so it won't be specific to DayDream.

liaxim:
The daydream-specifics could be encapsulated in a separate unit named clearly so there is no confusion about it. gvr-daydream is too generic of a name and could be misunderstood to mean a lot more than controller support - it is clear by now that pretty much nobody reads the documentation on github. Needs further clarification what to expect out of it - maybe a toast on app launch or something. Also I thought we want to keep the number of samples in check.

There are other options too - since there isn't much code involved (it looks like less than 10% from the code in the sample is about the controller) it could be an article in the official website. Once we have the abstraction we can showcase that as a sample too.

liaxim:
Should we merge or will we reconsider the approach?

rahul27:
Here is what we could do:

We could push it to a separate repo or my personal repo and point developers to that repo for reference while we work on the IO framework that supports sufficient abstraction to detect the gearvr touchpad or the daydream controller when available and handle them accordingly.

@thomasflynn what do you think?

rahul27:
Will push this to a private repo and update the link to the wiki.

PR #393, opened 2017-01-04
Add a buildGvrfAndDemos gradle property. When set it changes the

liaxim:
dependencies so the demos can be built together with the framework
and the backends.

Set buildGvrfAndDemos to true in gradle.properties and then add
the demo module to extra_settings.gradle; e.g.

include ':gvr-simplesample'
project(':gvr-simplesample').projectDir=new File('<insert-your-path-here/GearVRf-Demos/gvr-simplesample/app')

Do note that the "app" directory has to be referenced.

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -19,6 +19,8 @@ android {
 
     buildTypes {
         debug {
+            debuggable = true
+            jniDebuggable = true
             resValue 'string', 'app_name', System.getProperty("appName")
         }
         release {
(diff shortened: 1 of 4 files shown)


common.gradle
```diff
@@ -81,35 +83,61 @@ if(!rootProject.hasProperty("useLocalDependencies")) {
...
                 compile(name: 'backend_oculus-debug', ext: 'aar')
-            } else {
-                compile "org.gearvrf:backend_daydream:$gearvrfVersion"
-                compile "org.gearvrf:backend_oculus:$gearvrfVersion"
+                compile 'com.google.vr:sdk-base:1.10.0'
+                compile 'com.google.protobuf.nano:protobuf-javanano:3.0.0-alpha-7'
             }
+        }
+    }
+} else if(rootProject.hasProperty("buildGvrfAndDemos")) {
```
rahul27: Do we need another gradle property here?

Now that we have the builds up on jcenter/mavencentral, couldn't we just default to them if it cannot find the framework at the default relative path. Something along these lines:

https://github.com/rahul27/GearVRf-Demos/commit/3b4b2277e1bb4e913529eb80a38fcb2b0ee48362

It would make the builds easier without the need to provide the extra build properties needed for gradle. It also means we could get rid of the gearvrf-libs folder.

liaxim: I still prefer the aars since they save a good chunk of time when rebuilding the demos. The automated framework build produces aar artifacts. The automated demos build is set up to use those aars. One day we can refactor the thing. The goal of this pull request is just to make sure debugging works.

If you'd like you could propose a change in a pull request but we need to take into account the various use cases we have for the aars.

liaxim:
Cleaned everything and tried again. Verified that debugging works. If you pull this change in and set the property you must do a git clean -fdx and build everything before trying to debug.

common.gradle
rahul27: Reg. rebuilding if we chose to go with the local build option with the automated build, once the framework has been built it does not take much time between demos. Gradle does a good job of not rebuilding project when it has already been built once and no code changes have been made since. The time should be comparatively the same.

liaxim: Probably. But the demos build is working just fine right now. I really don't want to keep dealing with it. As I said, you could propose the changes across the board.

liaxim: If you strongly object to having this new property then I will not merge and we can continue discussing the issue offline.

rahul27: Think we should take this offline.

rahul27: This change can go in for now. We can sort out the build later.

PR #396, opened 2017-01-10
gvr-360video: increase the sphere radius

liaxim:
makes focusing better


liaxim:
Together with https://github.com/Samsung/GearVRf/pull/1011 I find the behavior of our app to be more or less identical to the Oculus 360 Video app.

PR #397, opened 2017-01-18
Updates to reflect the removal of getNextMainScene

liaxim:


liaxim:
Ready for review.

gvr-360video/app/src/main/java/org/gearvrf/gvr360video/Minimal360Video.java
```diff
@@ -33,7 +33,7 @@
     /** Called when the activity is first created. */
     @Override
     public void onInit(GVRContext gvrContext) {
-        GVRScene scene = gvrContext.getNextMainScene();
+        GVRScene scene = gvrContext.getMainScene();
```
rahul27: There might be a little confusion about the use of getMainScene(). Say I need to create an app with multiple scenes. 

With this model I would need to request the framework for the first scene and then create the next 2 scenes and then set them using setMainScene(). 

The call to getMainScene() in this case is just a special case where the framework creates the first scene for you and presents it to you the first time (making it more than a function that returns the current main scene). In this case, the user need to be educated not to create the first main scene but instead to request it from the framework.

I feel a cleaner approach would be if the user created all three scenes and set them one by one. 

Let me know what you think.

liaxim: With or without this change the model for apps with more than one GVRScene is not changed. If that has to happen, I rather have it as a separate pull request.

My thinking is to completely remove the capability to change or create a scene by end users. Changing the scene raises edge cases we deal with varying success to no clear benefit.

The real question is why would you need an app with 3 scenes. So far I have not seen a good reason to do so. gvr-immersivepedia does it but it can just as easily have its two scenes as regular scene objects in a gvrswitch node. Why support doing the same thing in more than one way?

NolaDonato: I agree that we should remove the ability to change the scene. You can still clear it and add different objects if you want to replace it so we are not actually removing functionality.

I think we could implement setMainScene by just copying the relevant properties from the input scene to the one and only main scene. Maybe we could rename it updateMainScene so it is clearer that we are not replacing it, but just changing it.

liaxim: @rahul27 How does it all sound? Should we post-pone merging this change?

rahul27: @liaxim if the plan is to only allow a single scene in the future then this change is fine.

liaxim: Thanks. This is the idea.

PR #398, opened 2017-01-25
Fix the lifecycle for gvr-sceneobjects

rahul27:
* Fixed onPause and onResume.
* Only release the camera onDestroy.

r.rudradevan@samsung.com


liaxim:
Per Camera API (https://developer.android.com/reference/android/hardware/Camera.html):
``Important: Call release() to release the camera for use by other applications. Applications should release the camera immediately in onPause() (and re-open() it in onResume()).``

The problem with onDestroy is that the Runtime doesn't guarantee it will be invoked.

rahul27:
The CameraSceneObject does not react well to onPause() and onResume. I wanted to change the sample without affecting the CameraSceneObject but if we are to release the camera on pause then we might have to change the way the CameraSceneObject works.

Let me see if I can fix it.

liaxim:
Looks good. Will merge Wednesday after lunch.

PR #402, opened 2017-02-14
Give .obj extension to meshes

NolaDonato:
New assimp relies on file extension to figure out what type of file it is. Earlier versions assumed the type as .OBJ for loadMesh. Now loadMesh can load other types but the filename must contain the proper extension.

Also replaced used of deprecated GVRContext.loadModel


liaxim:
Looks good. Will merge Wednesday morning if there are no comments.

PR #403, opened 2017-02-15
Add bare-bones template app that is self-contained and has minimalistic gradle files.

liaxim:
Should serve either as a reference or as a starter project for those who want to start an app from scratch.

Works with 3.1.


liaxim:
Will merge on Friday morning if there are no comments.

PR #416, opened 2017-03-07
Fix deprecated Apis in immersivepedia, keyboard, modelviewer2

NolaDonato:
Do not merge: these are my changes to the demos. Abhijit will be merging these changes into his pull request #410 to avoid confusion.

gvr-immersivepedia, gvr-keyboard: Use events generated by the picker instead of explicitly scanning the pick list each frame
gvr-keyboard: move skybox texture from drawable to res/raw, 
gvr-modelviewer2: use variable to avoid redundant indexing, fix build.gradle to look at buildGvrfAndDemos flag

GearVRF DCO signed off by: Nola Donato nola.donato@samsung.com

PR #419, opened 2017-03-16
Removing deprecated Apis from demos

NolaDonato:
1. Use picking events instead of scanning the list of picked items
2. Use the asset loader instead of the context
3. Use OBJ and MTL files instead of separately loading meshes and textures

This is a merged, squashed commit of Abhijits previous pull request #416 

GearVRF DCO signed off by: Abhijit Jagdale & Nola Donato apjagdale@samsung.com nola.donato@samsung.com


NolaDonato:
The next step for the demos is to use GVRShaderTemplate to author all the custom shaders. We are deprecating the other way. This is basically almost the same work as we did for eliminating the built-in shaders so we can probably pull the changes from that branch. This should be a separate pull request in case we don't want to hold up the release for it.

blur.txt
```diff
@@ -0,0 +1,66 @@
+//pref
```
liaxim: This file should probably be removed.

liaxim:
gvr-bullet: on collision the sphere's texture used to change to a blue one; no longer works.

liaxim:
There are few commented out lines. Please remove them.

liaxim:
I assume we want this for the 3.2 release?

thomasflynn:
I believe that's what @NolaDonato wanted.

liaxim:
gvr-keyboard: when you give a wrong answer previously the red sphere animated in and tracked the head orientation; no longer happening. Is this intentional?

liaxim:
So the only obvious difference, probably unintended, is with gvr-keyboard. The rest looks good and behaves the same as far as I can tell. Though immersivepedia may use another pair of eyes.

thomasflynn:
confirmed the issue with the keyboard.  @NolaDonato is going to look at it.  I also verified the functionality and performance of immersivepedia matches master.

NolaDonato:
I fixed the issue with gvr-keyboard

liaxim:
This might be nitpicking but the animation is not the same. It used to be pretty smooth and now is rather abrupt. @thomasflynn your call.

thomasflynn:
which animation?  the one that rotates around?  I can double-check it.
in general, the performance doesn't match what i remember, but the performance was the same on master.

liaxim:
Just give a wrong answer. A red sphere animates in then rotate your head around. It is clearly different from master.

liaxim:
I also noticed the selected key can get kind of stuck but it is the same in master. Performance-wise I have not noticed anything. Will double-check with much older version when we start the final release testing.

thomasflynn:
oh, you're talking about the gvr-keyboard problem?  I showed that to Nola last night.  I have not tried her fix yet.  i was referring to immersivepedia in my earlier comment.

liaxim:
Will check gvr-immersivepedia's performance.

liaxim:
Using the 3.1 release and the latest development snapshot yields identical performance for me. Rendering at or very close to 60fps according to the VrApi logs. What fps do you see when you find the performance not as good?

thomasflynn:
I wasn't looking at the logs.  I could see the cursor doubling and sometimes tripling...which usually indicates that frames are dropping.

thomasflynn:
for reference, i was running this on an Adreno S7 Edge.

NolaDonato:
Abhijit's fix still did not pass Tom's inspection :-( I poked at it more and now Tom passes it. I think we finally nailed this one.

apjagdale:
Great.  :-)

liaxim:
Looks good. Will merge Tuesday morning.

PR #420, opened 2017-03-17
Update the daydream library to v1.30.0

rahul27:
r.rudradevan@samsung.com

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -93,8 +93,7 @@ if(rootProject.useLocalDependencies) {
         if (rootProject.hasProperty("only_daydream")) {
             debugCompile(name: 'backend_daydream-debug', ext: 'aar')
             releaseCompile(name: 'backend_daydream-release', ext: 'aar')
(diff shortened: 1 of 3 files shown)


rahul27:
Depends on https://github.com/Samsung/GearVRf/pull/1092.

common.gradle
```diff
@@ -136,13 +132,11 @@ if(rootProject.useLocalDependencies) {
 
         if (rootProject.hasProperty("only_daydream")) {
             compile "org.gearvrf:backend_daydream:$gearvrfVersion"
```
liaxim: For the case where we use maven you could make the daydream backend's dependencies transitive:
```
            compile("org.gearvrf:backend_daydream:$gearvrfVersion") {
                transitive = true
            }
```
Then remove `` compile 'com.google.vr:sdk-base:1.30.0'``. This way at least one of the branches will no longer need updating.

liaxim:
Will merge Tuesday morning.

PR #424, opened 2017-03-21
Adding gvr-blurfilter gaussian blur filter in two passes

NolaDonato:
This sample tests render to texture and authoring post effect shaders. It requires GearVRF PR #1099 and PR #1098

Related to https://github.com/Samsung/GearVRf/issues/1060


liaxim:
The new sample doesn't work on my Mali Note 5 and S7. Error:
```
03-30 09:27:46.431 14399 14461 E gvrf    : Could not compile shader 35632:
03-30 09:27:46.431 14399 14461 E gvrf    : 0:3: P0003: Extension 'GL_OES_EGL_image_external_essl3' not supported
03-30 09:27:46.431 14399 14461 E gvrf    : 0:7: L0001: Expected token '{', found 'identifier'
03-30 09:27:46.431 14399 14461 E gvrf    : 
```

gvr-blurfilter/app/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+if(file("../../common.gradle").exists()) {
+    apply from: '../../common.gradle'
+}
+else if(file("../../../../GearVRf-Demos/common.gradle").exists()) {
+    apply from: '../../../../GearVRf-Demos/common.gradle'
+}
+
+android {
+    defaultConfig {
+        applicationId "org.gearvrf.test"
```
liaxim: Please fix the applicationId to match the one in AndroidManifest.

gvr-blurfilter/app/src/main/res/raw/gaussianblurhorz.fsh
```diff
@@ -0,0 +1,28 @@
+#extension GL_OES_EGL_image_external_essl3 : require
```
liaxim: I guess enabling both may be an option?
```
#extension GL_OES_EGL_image_external_essl3 : enable
#extension GL_OES_EGL_image_external : enable
```

Although after I did that I got ``0:21: S0001: No matching overload for function 'texture' found``.

NolaDonato:
I have asked Roshan to take care of this in the renderer. We had talked about doing this earlier but never actually got around to it. We should look for this extension before compiling the shader and choose the one appropriate for the device.

liaxim:
It is a problem in the demo right now. We could temporarily enable both extensions until there is something in GVRf that deals with it. Or whatever workaround seems appropriate. The fact is that the demo doesn't work on any of my Mali device.

Should the demo work on Mali devices if I replace GL_OES_EGL_image_external_essl3 with GL_OES_EGL_image_external?

liaxim:
Seems to me GL_OES_EGL_image_external means texture2D or texture2DProj have to be used instead of texture for sampling.

roshanch:
ready for review.

liaxim:
Looks good. It would be nice if there is more blur so you can immediately tell. My eyesight is blurry enough and I wasn't sure which is which until changing the offset in the shaders.

Planning to merge Monday morning.

PR #426, opened 2017-03-22
Modified demos to support the daydream 64 bit backend

rahul27:
r.rudradevan@samsung.com

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -42,11 +42,11 @@ android {
         }
     }
 
-    // ignore the x86 and arm-v8 files from the google vr libraries
+    // ignore the x86 files from the google vr libraries
     packagingOptions {
+        exclude 'lib/x86/libgvr.so'
         if (!rootProject.hasProperty("ARM64")) {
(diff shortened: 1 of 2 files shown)


rahul27:
Depends on https://github.com/Samsung/GearVRf/pull/1105

thomasflynn:
question:  do we need to have the 'ARM64' property or do any excluding of .so's now?  seems we can include everything all the time now and let the platform decide which libraries to use.

liaxim:
If android sees any 64bit binary it will assume the app is 64bit. Not always the case but since we are shipping a library not really our responsibility. For the demos build it is useful to have an easy way to force an architecture.

So we could include everything in our aars but we'll have to check the extensions too. Not sure what the deal will be with bullet and especially gdx.

liaxim:
Planning to merge Thursday afternoon.

liaxim:
Will need a rebase

rahul27:
rebased.

PR #428, opened 2017-03-24
gvr-sample: update so capturing works as expected

liaxim:


liaxim:
Complements https://github.com/Samsung/GearVRf/pull/1110

liaxim:
With frustum culling on, some things get culled completely and all you get is a black screen.

PR #432, opened 2017-03-30
Use mavenCentral dependency for gson

liaxim:


template/GVRFApplication/app/build.gradle
```diff
@@ -41,6 +41,7 @@ ext.gearvrfVersion='3.1.2-SNAPSHOT'
 project.ext.jomlVersion = "1.9.1-SNAPSHOT"
 
 dependencies {
+    compile 'com.google.code.gson:gson:2.7'
```
rahul27: This line requires the snapshot build to include the framework change. We might need to build a new snapshot immediately after merging it.

liaxim: Snapshot updated.

PR #435, opened 2017-04-07
gvr-camera2renderscript: fix crash on S8

liaxim:
Looks like you can tell at runtime whether the yuv-rgb conversion is necessary (it is for all recent phones apparently). I can't find a phone that doesn't need the conversion so can't verify. I probably need a Note4 with Android L. KitKat just doesn't work due to the use of 21 apis. For now using the old and tried method.

---


liaxim:
Will merge Wednesday morning.

PR #439, opened 2017-04-13
gvr-modelviewer2: Fix the build from maven

liaxim:
+ remove unnecessary property from gradle.property

Needed for the 3.2 release

Removing the org.gradle.parallel allows setting it in the global gradle.properties to true. Building in parallel seems to work pretty good nowadays and there is a great speedup when building and installing the demos.


liaxim:
Will merge today after lunch.

PR #447, opened 2017-05-02
Update MainActivity.java

mohsinjuni:
Setting scene's background color to WHITE. This would help a new user or developer to easily view the output window.


liaxim:
Any color other than white or black might be preferable since you may end up stuck with either on Android, depending on the circumstances.

PR #450, opened 2017-05-23
Apply the scale to the video, not the sphere.

thomasflynn:
Setting the scale on the sphere tells the renderer what matrix to apply
when rendering that SphereSceneObject.  however, we're not using the
SphereSceneObject, but rather the vertices from the sphere.  So, apply
the scale to the VideoSceneObject instead.

tom.flynn@samsung.com


liaxim:
For issue https://github.com/Samsung/GearVRf/issues/1205

PR #455, opened 2017-06-08
Set the minimum support API level to 21

liaxim:
Depends on https://github.com/Samsung/GearVRf/pull/1230


thomasflynn:
tested this along with the associated with framework pull-request. ran on a Note4 with Lollipop. seems to be working just well.

PR #459, opened 2017-06-20
Test many video scene objects

joelsung:
I'm trying to play multiple videos in Gear VR, and tested 360Video from manyvideos.
But I'm having trouble with Error "Configuration with the name 'default' not found"

I also checked previous issue, but couldn't resolve the problem by replacing GVRMain/setMain with GVRScript/setScript.
https://github.com/Samsung/GearVRf/issues/609


liaxim:
@joelsung I updated the branch. It builds and runs now. You may have to refresh your gradle dependency cache.

Eventually we will have to close this pull request since this is not a change that belongs to master.

joelsung:
Thanks a lot!!

PR #462, opened 2017-06-22
gvr-immersivepedia: use the app theme that removes the white screen that flashes at launch

liaxim:


liaxim:
Planning to merge on Friday.

PR #468, opened 2017-06-28
Fix EGL_image_external

dnvasilev:
Different phone models need different extensions for samplerExternalOES:
Adreno-based, GL_OES_EGL_image_external_essl3
Mali-based, GL_OES_EGL_image_external
Instead of 'require' a particular extension in the fragment shader, 'enable' both.


liaxim:
Looks good to me. Tried to bring this up in the past.

liaxim:
Planning to merge next week

PR #472, opened 2017-07-12
Fix gvr-shadows

dnvasilev:
New API does not support shadow-near and shadow-far.


NolaDonato:
shadow_near and shadow_far are being eliminated. The correct way to set the shadow near and far planes is to modify them on the shadow camera. GVRLight.setShadowRange does this.

PR #475, opened 2017-07-19
gvr-immersivepedia: convert shaders to 300 and enable multiview support

liaxim:


liaxim:
@roshanch Please review. Also enables multiview for gvr-immersivepedia. I think it is time to enable it by default for select apps.

roshanch:
looks good to me.

liaxim:
Planning to merge on Thursday.

PR #477, opened 2017-07-24
fix for outline-sample for vulkan

roshanch:


roshanch:
I modified gvr-cubemap to remove extra scene-object as compared to master. This sample has gvrStats enabled which uses external textures. To run it on vulkan comment that line.

liaxim:
Shouldn't the stats be handled a bit more gracefully by the vulkan renderer? Like ignoring them automatically.

NolaDonato:
The stats don't work because of an unsupported feature. We can do this:
if (!GVRShader.isVulkanInstance()) { add the stats... }

liaxim:
Should be handled by the framework, not the app, I assume.

NolaDonato:
Actually, we will eventually get these features working in Vulkan and then we will remove all this stuff from the demos.

dnvasilev:
Needs rebase

PR #478, opened 2017-07-24
add gvr-rendertotexture

sushantojal:
Adding "gvr-rendertotexture" demo to this branch since it was present in master but not in this branch.
works with openGL.
does not work with vulkan since this demo has lighting.


NolaDonato:
Edit the demo to not add the light in the Vulkan case so we can test RTT:
if (!GVRShader.isVulkanInstance()) { addLight(...); }

sushantojal:
RTT is not supported on the vulkan branch right now, so this demo will work only with openGL on the vulkan branch.

NolaDonato:
Roshan is fixing this.

dnvasilev:
@NolaDonato : should I merge?

PR #480, opened 2017-07-27
gvr-cubemap: reflective sphere is back

dnvasilev:
Fix: reflective sphere was removed by accident


liaxim:
I don't see it. Needs the attached diff to show up: 
[diff.txt](https://github.com/gearvrf/GearVRf-Demos/files/1184272/diff.txt)

Then again why do we need it? Has been removed ages ago, at times when the demos lived in the GearVRf repo. Commit f6a7d7d2f, no hint of a reason.

We copied files around couple of times for whatever reasons. So tracing the history can be annoying.

liaxim:
@dnvasilev Please let me know does the diff work for you? The sphere doesn't show for me otherwise.

liaxim:
Planning to merge on Friday. Then will create a pull request to really show it.

PR #481, opened 2017-07-27
Vulkan: eliminate attachEyePointeeHolder calls in gvr-controls

NolaDonato:
We will be deprecating this function


NolaDonato:
wrong branch

PR #488, opened 2017-08-03
gvr-cubemap, gvr-360video: let these app pick up assets from the sdcard

liaxim:
gvr-360video: tries to first play /sdcard/gvr-360video/video.mp4 before falling back to the one in the assets

gvr-cubemap: displays all pngs and bmps from /sdcard/gvr-cubemap/ as a full screen quad with the camera frozen
gvr-cubemap: displays all zips from /sdcard/gvr-cubemap/ as cubemaps with a free camera

+ use the GVRfAppTheme style to avoid white screen flashing on startup
+ pause/resume the video in gvr-360video on tap


liaxim:
Planning to merge this on Friday after lunch.

PR #492, opened 2017-08-05
Fixed gvr-modelviwer2 on vulkan branch

apjagdale:
Requires #1365


NolaDonato:
I fixed a crash in BoundingBoxCreator caused by not enough texcoords in the array (which 4.0 checks for but 3.2 did not)

NolaDonato:
ModelViewer2 crashes on Vulkan when you try to display the bear. To reproduce copy the bear and its textures from https://github.com/NolaDonato/GearVRf-Tests/tree/master/jassimp/animals to the GVRModelViewer2 directory on the phone. The model viewer will try to import it from this directory. It crashes in adjustRenderingOrderForTransparency (only on Vulkan)

PR #499, opened 2017-08-14
Vulkan: gvr-events: get event hit point properly

NolaDonato:


apjagdale:
Demo works as expected. But the Android frame flickers at random intervals.

PR #516, opened 2017-09-19
Modifying gvr-controller demo, adding gvr-keyboard

NolaDonato:
The gvr-controller demo now supports either Gear or Gaze controller and shows how to use the new controller APIs for testing connectivity.
I am adding Ragnar's keyboard demo because I tried to make it use the new touch events and picker APIs but it has a problem still - keyboard disappears on key up

---
Required for https://github.com/Samsung/GearVRf/pull/1455


gvr-controller/app/src/main/java/org/gearvrf/sample/controller/SampleMain.java
```diff
@@ -173,162 +184,241 @@ public void onInit(GVRContext gvrContext) {
...
-                cursor.getRenderData().setRenderingOrder(100000);
-                controller = new GVRGearControllerSceneObject(mGVRContext);
-                controller.setCursorController(gvrCursorController);
-                mPicker.setPickRay(0.0f, 0.0f, 0.0f, 0.0f, 0.0f, DEPTH);
-                controller.setRayDepth(DEPTH);
-                controller.setCursor(cursor);
-                controller.attachComponent(mPicker);
-                gvrCursorController.addControllerEventListener(controllerEventListener);
-            }else{
+        public void onCursorControllerAdded(GVRCursorController gvrCursorController)
```
liaxim: The old api was already making users do too much. On the surface at least, it now looks the users have even more work to do. Can't all these ifs be pushed into the framework?

gvr-controller/app/src/main/java/org/gearvrf/sample/controller/SampleMain.java
```diff
@@ -173,162 +184,241 @@ public void onInit(GVRContext gvrContext) {
...
+            GVRCursorControllerSceneObject controllerModel = new GVRCursorControllerSceneObject(mGVRContext);
+            mainScene.getMainCameraRig().getHeadTransformObject().addChildObject(controllerModel);
+            controllerModel.setCursorController(controller);
+            controllerModel.setRayDepth(DEPTH);
+            controllerModel.setCursor(cursor);
+            controllerModel.enableSurfaceProjection();
+            controller.addPickEventListener(mPickHandler);
+            mControllerModel = controllerModel;
+            mCursorController = controller;
+            mGazeController = (GVRGazeCursorController) controller;
```
liaxim: Why is the abstraction of the past removed? The users need to be aware of the specific instances of cursor controller classes now it seems. Presumably the users need to do different things depending on the type. We should be going into a direction of abstracting more than the old api did instead.

liaxim:
Could be the app, could be the fw, can't tell but with this version, if you select an object and keep it selected by holding the button, it can get randomly unselected. Seems to have something to do with going over other objects in the scene.

gvr-controller/app/src/main/java/org/gearvrf/sample/controller/SampleMain.java
NolaDonato: The 3D Cursor extension attempts to solve this problem. That is our current method of pushing these ifs into the framework. It supports configuration files and the idea of multiple active cursors.

gvr-controller/app/src/main/java/org/gearvrf/sample/controller/SampleMain.java
NolaDonato: This has always been true. This sample has never supported the Gaze controller before. The Gaze controller wasn't able to generate touch events by itself before my change. The app had to intercept them. There have always been a bunch of different cursor controller classes - I did not change these APIs.

gvr-controller/app/src/main/java/org/gearvrf/sample/controller/SampleMain.java
liaxim: gvr-controller does not use the 3dcursorlibrary. All I can say that now there is even less abstraction than before, without the (flawed) 3dcursorlibrary..

gvr-controller/app/src/main/java/org/gearvrf/sample/controller/SampleMain.java
liaxim: Then we should do correctly the gaze controller and keep it transparent to the end users.

liaxim:
I expect some of my comments don't make sense but I will spend more time with these changes and try to provide more constructive feedback where I also suggest solutions. All I care about is that we at least keep the existing level of complexity in the public APIs.

liaxim:
gvr-keyboardview shows no keyboard for me on s8. Does this app have to be part of the pull request?

ragner:
@liaxim,
gvr-keyboardview is missing a GVRPicker since GVRPickerInput class was removed.

Today I will try fix it with latest  "controller_reform"

ragner:
Hi @all,
I am facing the following issue at gvr-keyboardview with current "controller_reform"

```

E/AndroidRuntime: FATAL EXCEPTION: GLThread 14802
                                                   Process: org.gearvrf.keyboardview, PID: 19560
                                                   java.lang.NullPointerException: Attempt to invoke virtual method 'void org.gearvrf.io.GearCursorController.attachReader(org.gearvrf.io.GearCursorController$ControllerReader)' on a null object reference
                                                       at org.gearvrf.OvrViewManager.onSurfaceCreated(OvrViewManager.java:317)
                                                       at org.gearvrf.OvrVrapiActivityHandler$8.onSurfaceCreated(OvrVrapiActivityHandler.java:374)
                                                       at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1530)
                                                       at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1262)
10-02 03:55:51.945 19560-19560/? D/ViewRootImpl@fdb6400[MainActivity]: MSG_RESIZED_REPORT: frame=Rect(0, 0 - 2960, 1440) ci=Rect(0, 0 - 192, 0) vi=Rect(0, 96 - 0, 0) or=2
```

NolaDonato:
Are you using the gvr-keyboardview from PR #516 in GearVRf-Demos? I do not get this exception when I run gvr-keyboardview.

ragner:
Hi @NolaDonato,
yes I am using the gvr-keyboardview from  #516. But it looks to be a issue from the framework.

I don't know why  mInputManager.getGearController() is returning a null value: 

```
 void onSurfaceCreated() {
        super.onSurfaceCreated();
        mRotationSensor.onResume();
        mGearController = mInputManager.getGearController();
        mGearController.attachReader(new OvrControllerReader(mActivity.getActivityNative().getNative()));
    }
```

ragner:
I can reproduce this issue every time.

You just need launch a application with a empty onInit():

  `public void onInit(final GVRContext gvrContext) {}`

It looks to be a concurrency issue caused by asynchronous tasks ...

NolaDonato:
How do you get gvr-keyboardview to work with an empty onInit()? There is no GVRContext...

ragner:
I make this change just to have a simple way of reproduce the issue.
Keyboard does not work for sure.

ragner:
@NolaDonato, after your last commits it is working well now ...

It looks that create a new instance of GVRPicker is not necessary anymore!

PR #524, opened 2017-10-04
adding google vr dependency

roshanch:


liaxim:
@roshanch Can we close?

PR #525, opened 2017-10-04
Adapt gvr-events for the new controller support

liaxim:


liaxim:
@NolaDonato Please advice whether this is the correct approach. Thanks.

While it responds to touch events, there is that:
1. event latency; there seems to be noticeable delay between action_down and action_up
2. same cursorcontroller instance registered twice; getting this back from the framework
3. not tested with gearvr's touchpad
4. scrolling the listview doesn't work

NolaDonato:
The cursor controllers are registered each time the main scene changes. If you are using a splash screen, it will register them for that and then again when the scene changes. The application used to have to scan the controllers, I tried to make it automatic. I could suppress the second call by having input manager remember whether it has scanned or not.

liaxim:
Waiting for Nola controller updates

PR #528, opened 2017-10-06
Fixing gvr-shadows

apjagdale:


gvr-shadows/app/src/main/java/org/gearvrf/gvrshadowssample/ShadowsMain.java
```diff
@@ -99,8 +78,8 @@ private GVRSceneObject createDirectLight(GVRContext context)
         GVRDirectLight light = new GVRDirectLight(context);
 
         light.setCastShadow(true);
-        light.setShadowRange(1.0f, 150.0f);
         lightNode.attachLight(light);
+        light.setShadowRange(1.0f, 150.0f);
```
liaxim: Does the order of invocation matter (setShadowRange)?

liaxim:
@apjagdale Isn't the stormtrooper supposed to cast a shadow?

liaxim:
Well, the shadow is there now but on couple of occasions it wasn't. Will merge on Wednesday.

PR #530, opened 2017-10-09
Fixing Scene Item issue which dissappears onFocus

apjagdale:


liaxim:
Will merge on Wednesday

PR #532, opened 2017-10-09
Fixing gvr-Javascript

apjagdale:
Modifying shaders.

### DIFF
--- a/gvr-javascript/app/src/main/res/raw/fragment.fsh
+++ b/gvr-javascript/app/src/main/res/raw/fragment.fsh
@@ -12,13 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#extension GL_ARB_separate_shader_objects : enable
+#extension GL_ARB_shading_language_420pack : enable
 precision mediump float;
-layout (std140) uniform Material_ubo
-{
-    vec4 u_color;
-};
-
-out vec4 fragColor;
+@MATERIAL_UNIFORMS
+layout (location = 0) out vec4 fragColor;
 void main() {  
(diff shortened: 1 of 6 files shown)


gvr-javascript/app/src/main/res/raw/vertex.vsh
```diff
@@ -11,25 +11,14 @@
...
-     mat4 u_mv_it;
- #endif
-     mat4 u_model;
-     mat4 u_view_i;
-     float u_right;
-};
+precision highp float;
+@MATRIX_UNIFORMS
+
+layout(location = 0) in vec4 a_position;
```
liaxim: Does this mean the gvr-accessibility no longer will work with multiview enabled?

gvr-javascript/app/src/main/res/raw/fragment.fsh
```diff
@@ -12,13 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#extension GL_ARB_separate_shader_objects : enable
```
liaxim: These extensions are in every shader in the framework. Are they not something that the framework should be adding since it expects them apparently? When using OpenGL their removal has no effect. If only Vulkan requires them then there is even a stronger case for this to be handled by the fw.

liaxim:
Will merge on Wednesday.. Though

- occasionally the text size is rather larger; it was there in the past though much more infrequent
- on tap the text color should be animated; in the past occasionally it didn't work but now it never works

apjagdale:
@liaxim For running the demo in multiview it requires #PR1520 to be merged

PR #538, opened 2017-10-13
Fixing javascript demo

roshanch:
disables multiview default. modifying shaders. It runs on vulkan now.


liaxim:
Why multiview needs to be disabled by default?

roshanch:
@liaxim otherwise it wont' work with vulkan.

liaxim:
Can't multiview enabled be just ignored for vulkan?

Like the lights.. The first thing that needs to go in the next release is the public isVulkanInstance method.

roshanch:
we do have isVulkanInstance API. I will add that for multiview. But for this PR only change needed in xml.. shaders should have capability to switch to non-multiview. #ifdefs

liaxim:
Actually what I implied is that the framework should be taking care automatically of all these things. Apps should not need to check renderer instances or be aware of when to enable or disable multiview. They should be agnostic and the framework should be making the decisions for them. Otherwise it can work but .. most users would not appreciate the complexity and all the tinkering that it takes.

So for next release we should not have public APIs that talk about render instances or multiview. May not be achieved but must be a goal we strive for.

From what I've seen so far nobody is using multiview. Because it is disabled by default and comes with too many caveats and looks like even more in 4.0.

PR #539, opened 2017-10-13
gvr-cardboard-audio: fix for this release by removing the light

liaxim:


liaxim:
Looks now the same as in maint. The demo in maint is in need of updating to properly use lights in the maint branch but that is not all. There is something wrong in the master framework when the light is in there and this model is used. There is no time to properly investigate and will come back to it in the future.

On an unrelated note the short and long back key press work as expected. Not sure what fixed it.

PR #540, opened 2017-10-16
gvr-renderableview: fix android views use; prevents a white flash at startup

liaxim:


ragner:
Looks good to me.

liaxim:
@ragner Thanks for reviewing. Planning to merge on Monday.

PR #541, opened 2017-10-16
Adding precision to work on s7

apjagdale:
Depends on #PR1550.
Fixes issue on Mihail's phone.


liaxim:
But why did it work previously with medium?

PR #542, opened 2017-10-20
gvr-immersivepedia: fix interference by the gear controller by disabling the gear controller object

liaxim:
- the gear controller objects has a picker of its own that would generate pick events that set the app's picker handler state incorrectly
- also remove redundant PickerHandler (which oddly enough wasn't getting gc-ed which I will look into in the future)

---
Depends on https://github.com/Samsung/GearVRf/pull/1557

### DIFF
--- a/gvr-immersivepedia/app/src/main/java/org/gearvrf/immersivepedia/Main.java
+++ b/gvr-immersivepedia/app/src/main/java/org/gearvrf/immersivepedia/Main.java
@@ -15,10 +15,13 @@
 
 package org.gearvrf.immersivepedia;
 
+import android.media.MediaPlayer;
+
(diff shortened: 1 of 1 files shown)


liaxim:
- The cursor controllers should be disabled by default but it is something we can do in master. For now doing this app-specific hack.
- Not convinced that sending picking events globally by all active pickers is that good of an idea; we could add a listener to the specific GVRPicker instance and be done with it. It is so easy now to have interference by other erroneous or not pickers.

NolaDonato:
I agree we should disable the controllers by default and make the app enable the one(s) it wants.
We can remove the events being sent to the scene now that the picker has its own listeners.

liaxim:
Planning to merge on Tuesday

PR #545, opened 2017-11-07
Update tutorials and samples for 4.0 API

nitosan:
1. Remove deprecated API usage
2. Simplify the samples
3. Separate Mesh loading tutorial with Animation tutorial


NolaDonato:
Why are you loading the mesh and texture separately in the tree animation tutorial?  I believe that FBX  file has the textures in it already. You can just load the model and start the animations with the START_ANIMATIONS import setting.

nitosan:
Yes, I tried loading fbx with texture, unfortunately found out that all FBX in this repo contains absolute path to the texture, result in the texture will fail to load in most cases. It also didn't help that if you load FBX directly, Gvrf will add additional `SceneObject` as root, so it's hard to attach material later.

So I think load them separately should be the recommended method. While I'm working on a section about how to export fbx correctly.

NolaDonato:
Loading them separately only works if you have a single mesh in the file. It fails for more complex hierarchies. I will get you a Trex model with good paths and look into changing the asset loader to fix bad paths.

tutorials/tutorial_3_model_animation/app/src/main/java/com/example/org/gvrfapplication/MainScene.java
```diff
@@ -19,31 +19,35 @@
...
-        character.getTransform().setRotationByAxis(45.0f, 0.0f, 1.0f, 0.0f);
-        character.getTransform().setScale(3, 3, 3);
-        character.getTransform().setPosition(0.0f, -0.4f, -0.5f);
-        gvrContext.getMainScene().addSceneObject(character);
-
-        final GVRAnimator animator = (GVRAnimator)character.getComponent(GVRAnimator.getComponentType());
-        animator.setRepeatCount(-1);
-        animator.setRepeatMode(GVRRepeatMode.REPEATED);
-        animator.start();
+//        GVRModelSceneObject character = gvrContext.getAssetLoader().loadModel("astro_boy.dae");
```
liaxim: Is this intentionally commented out?

liaxim:
tutorial_5_controller doesn't compile for me.

liaxim:
Planning to merge on Monday. I guess you are planning to address some things in the future.

PR #546, opened 2017-11-13
gradle: enable parallel gradle build

liaxim:


liaxim:
AS 2.3.3 didn't work with this. 3.0 does. It helps significantly when using the top-level project in AS.

liaxim:
Planning to merge on Wednesday

PR #547, opened 2017-11-14
gvr-immersivepedia: do not kill the app process from onDestroy

liaxim:


NolaDonato:
This fixes the crash on startup under VRTop for me.

PR #548, opened 2017-11-16
Modified demos to add new controller support

NolaDonato:
Modified demos to support Gear controller by using new controller API GVRInputManager.selectController. Heavily modified 3D cursor demos to actually use the features in the 3D cursor library instead of duplicating them.

This PR require GearVRf PR #1628.

Each demo is a separate commit. The demos changed are: 
gvr-3dcursor
gvr-3dcursor-simple
gvr-keyboardview
gvr-simplephysics
gvr-controls
gvr-immersivepedia
gvr-events
gvr-controller
gvr-accessibility


gvr-accessibility/app/src/main/assets/gvr.xml
```diff
@@ -28,9 +28,10 @@
         framebufferPixelsHigh="DEFAULT"
         framebufferPixelsWide="DEFAULT"
         showLoadingIcon="true"
-        useMultiview="true"
+        useMultiview="false"
```
liaxim: Why multiview needs to be turned off?

NolaDonato: It doesn't. I will turn it back on. I was testing both cases.

liaxim:
gvr-accessibility: enable all options on the circular menu near the bottom; for me when I click the Zoom out menu, the app registers Volume down action. Most if not options from this menu do not do what you would expect. Tested on S8.

liaxim:
gvr-simplephysics: not sure it is playable with the controller the way it is or maybe it is just me.

liaxim:
gvr-keyboardview: keyboard view clearly shows the picking problem apparent in most other apps

liaxim:
gvr-3dcursor: it just doesn't do what it used to do; can't select anything with a touch event. All my tests btw were in developer mode without headset.

liaxim:
gvr-3dcursor-simple: the controller is there unlike gvr-3dcursor; not quite sure you can do the same things you could in the past but have not tested with the version without the changes.

PR #555, opened 2017-11-30
gvr-javascript: comment out setSensor call that needs replacing

liaxim:
Will be properly addressed in the near future


liaxim:
Otherwise the row boat animation is not played

PR #556, opened 2017-11-30
Fixed demos to work with latest controller fixes

NolaDonato:
1. GVRInputManager.selectController remove context argument
2. Remove activity touch handler with Gaze controller - no longer needed
3. Improved event dispatch for demos based on widget plugin (scrolling
not working)

GearVRF DCO signed off by: Nola Donato nola.donato@samsung.com

Framework pull request: https://github.com/Samsung/GearVRf/pull/1646

PR #559, opened 2017-12-13
gvr-modelviewer2, gvr-widgetviewer: updates so events work after pause/resume

liaxim:
Depends on https://github.com/Samsung/GearVRf/pull/1669 and includes the changes from #556

522fdb5 - the one and only new commit for this change

---


NolaDonato:
I have tested this with modelviewer2 and widgetviewer and it works. I have included it in PR #556

PR #564, opened 2017-12-17
gradle version

pnemonic78:
Android gradle build tools plugin version 3.0.1 - requires minimum build tools 26.0.0

PR #574, opened 2017-12-29
Revert "gradle version"

NolaDonato:
Reverts gearvrf/GearVRf-Demos#564

creates problems with Nola's machine.  temporarily reverting.


thomasflynn:
probably not necessary to merge this.  turns out reverting the change in common.gradle was enough to unblock her.  we should figure out what's up with her system instead.

PR #578, opened 2018-01-11
gvr-lua: delete the sample; lua support has been removed

liaxim:


liaxim:
This is for the maint_v3.x branch

liaxim:
Merging. The demos build fails due to the now missing lua.

PR #580, opened 2018-02-05
gradle: rework the way backends are specified

liaxim:
Replace the archaic way of specifying the backends to use with, e.g.:

```
		backend_oculus=false
		backend_daydream=false
		backend_monoscopic=true
		backend=false
```

If none specified defaults to oculus+daydream, as before.

``no_oculus, only_daydream`` are being removed.


liaxim:
Also combinations are possible; if you'd like you could have all the backend included.

liaxim:
Travis job will be updated to specify backend_daydream=true after merging

liaxim:
Planning to merge on Wednesday

PR #582, opened 2018-02-06
Fix couple of app's package names

liaxim:


liaxim:
For the maint_v3.x branch

PR #584, opened 2018-02-07
gradle: add abiFilter so we can use the android profiler

liaxim:


liaxim:
@thomasflynn This fixes your problem after enabling the Android profiler. Now only the correct libsupportjni.so will be included in the apk.

PR #593, opened 2018-04-04
gvr-complexscene: make it more complex

liaxim:


liaxim:
Brings down the drawing fps to less than 30 on newer phones

PR #595, opened 2018-04-09
build: use shared std library

liaxim:
Requires https://github.com/Samsung/GearVRf/pull/1824


liaxim:
Planning to merge on Monday

PR #596, opened 2018-04-10
gvr-modelviewer2, gvr-widgetviewer: tweaks to make them work better

liaxim:
Set widget plugin's view size to fixed values. It doesn't have to match the display resolution. Since there are some absolute values in the gdx layout it makes things more stable.

Remove unused thumbnails-related code from gvr-widgetplugin.

Have gvr-widgetplugin call updateState after onInit to ensure the model is positioned as expected immediately after launch.

Hide the second widgetplugin pane (on the right side) of gvr-widgetviewer since there is nothing being actually drawn there. The demo needs a good amount of work to actually put the second pane to use.

### DIFF
--- a/gvr-modelviewer2/app/src/main/java/org/gearvrf/modelviewer2/ModelViewer2Activity.java
+++ b/gvr-modelviewer2/app/src/main/java/org/gearvrf/modelviewer2/ModelViewer2Activity.java
@@ -16,7 +16,6 @@
 package org.gearvrf.modelviewer2;
 
 import android.os.Bundle;
-import android.util.DisplayMetrics;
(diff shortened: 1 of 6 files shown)


Caprica666:
I tried both modelviewer2 and widgetviewer demos. Both show black screens with the latest master. I am using the S8 with Android 8. That might be one of the issues.

liaxim:
Just with the latest master or the latest master with this pull request in?

NolaDonato:
with your pull request. I will test with another phone too.

NolaDonato:
This works on Android 7. I think it is an Android 8 problem.

PR #602, opened 2018-04-30
gradle: support including build.gradle scripts from other projects

liaxim:
Makes our build system more flexible


liaxim:
Planning to merge on Thursday

PR #608, opened 2018-05-11
fixed bike in widget demo

NolaDonato:
new assimp imports it as 6 meshes instead of 1


liaxim:
Do you know why 6 instead of 1?

Planning to merge on Monday.

NolaDonato:
because now it makes a separate mesh for every separate face group in the OBJ file. This OBJ file has vertices, faces, vertices, faces ... Old assimp crammed all these into one mesh, new assimp makes six meshes. I tried the OPTIMIZE_GRAPH and OPTIMIZE_MESHES options and it still did that (these are supposed to flatten the hierarchy).

PR #610, opened 2018-05-18
Removed explicit mesh collider creation

crbozz:
Now that Physics loader API creates missing colliders (using mesh
collider) the explicit creation and attachment could be removed and
fewer lines of code are required.

Depends on: https://github.com/Samsung/GearVRf/pull/1876


crbozz:
Now depends on https://github.com/Samsung/GearVRf/pull/1885.

PR #612, opened 2018-05-24
Bugfix: physics were not working for loaded objects

crbozz:
The problem occurred after commit https://github.com/Samsung/GearVRf/commit/3c4d773d8a06ace028e2c95b0316e8ffd5856da6. Now loaded objects are removed from model scene object and added directly to scene root.


crbozz:
I am closing this pull-request because I am working on some other improvements to physics loader.

PR #615, opened 2018-05-31
use gltf models in gvr-widgetviewer

sushantojal:


sushantojal:
have left the readme files, since they have license info + credentials.
also have included all the gltf files associated with every model, for future testing.

apjagdale:
Changing light color has no effect on current model but the next ones.

liaxim:
If you click somewhere the light changes for the current model. @sushantojal Are you planning to fix it as part of this pull request?

sushantojal:
I'll take a look at it today, although the issue seems to be present with the dropdown-menu widget before this PR.

sushantojal:
Looks like this demo uses GL tex ids returned by the widget plugin, which cannot be used with vulkan. So although PBR works for vulkan, this demo specifically will not work. [ PR #1891](https://github.com/Samsung/GearVRf/pull/1891) in the framework can be merged. 
We can always add the vulkan instance check in the demo to display a model without the widget pane, but I think that check shouldnt be there in a demo app.

liaxim:
Planning to merge on Tuesday.

PR #617, opened 2018-06-08
Rag doll demo to multiple constraints

ragner:
This demos is a test case for https://github.com/Samsung/GearVRf/pull/1893

It is still in work in progress.


ragner:
I will do a new PR

PR #620, opened 2018-06-19
Add a listener for hover event

acosta:
Add a listener in gvr-events demo to test the hover support which is being proposed in GVRf (see https://github.com/Samsung/GearVRf/pull/1912).


liaxim:
Planning to merge on Wednesday.

PR #630, opened 2018-07-16
New video player demo

acosta:
This demo consists in a video player which has the following features:
- Use ExoPlayer (https://google.github.io/ExoPlayer/)
- Plays external (using DASH) and local videos
- Plays 360 videos
- Reposition the player on the scene
- User interaction to show video's info (using hover), tap in timeline, etc.
- Organize videos in a gallery with thumbnails
- Use Android layouts to create the screens


liaxim:
Looks pretty good

thomasflynn:
Overall, looks good, but I have one bit of feedback so far:

No matter where you are, pressing the 'back' button on the headset brings up the dialog to quit the application.  I would expect that if you were playing a video, it would back out of the video and go back to the grid.  And if you were looking at the grid of videos, it would back out to the top level.  

In short, I would expect to bring up the 'quit application' dialog only at the top level of the application.

acosta:
Thank you @liaxim and @thomasflynn for your feedback. We will work on this back button behavior.

thomasflynn:
thank you, much better now.

gvr-videoplayer/app/src/main/java/org/gearvrf/videoplayer/provider/asyntask/AlbumAsyncTask.java
```diff
@@ -0,0 +1,126 @@
...
+
+        String[] projection = new String[]{
+                VideoColumns._ID,
+                VideoColumns.BUCKET_DISPLAY_NAME,
+                VideoColumns.TITLE,
+                VideoColumns.DATA,
+                VideoColumns.DURATION,
+                VideoColumns.WIDTH,
+                VideoColumns.HEIGHT,
+                "is_360_video"
```
liaxim: Where does is_360_video come from? On Oculus GO there are frequent crashes because this column doesn't exist.

dsazulay: Samsung has its own media provider tables with some more fields besides the default of android.

liaxim: Oculus GO doesn't have these so the app needs to be able to tolerate it.

dsazulay: It can be removed from the app. The application has some other methods to detect 360 videos.

PR #631, opened 2018-07-24
Using GVRf script extension

xcaostagit:
These changes are necessary to use the new script extension
(c.f https://github.com/Samsung/GearVRf/pull/1950) for GearVRf.

### DIFF
--- a/gvr-javascript/app/build.gradle
+++ b/gvr-javascript/app/build.gradle
@@ -26,3 +26,8 @@ android {
         jniLibs.srcDir 'libs' // pre-compiled libraries
     }
 }
+
+dependencies {
+    debugCompile(name: 'script-debug', ext: 'aar')
+    releaseCompile(name: 'script-release', ext: 'aar')
+}
--- a/gvr-javascript/app/src/main/java/org/gearvrf/sample/gvrjavascript/GearVRJavascriptActivity.java
+++ b/gvr-javascript/app/src/main/java/org/gearvrf/sample/gvrjavascript/GearVRJavascriptActivity.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 
 import org.gearvrf.GVRActivity;
 import org.gearvrf.GVRAndroidResource;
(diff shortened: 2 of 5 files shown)


gvr-javascript/app/build.gradle
```diff
@@ -25,4 +25,9 @@ android {
         jni.srcDirs = [] // no auto generation of Android.mk
         jniLibs.srcDir 'libs' // pre-compiled libraries
     }
+}
+
+dependencies {
```
liaxim: Please update to
```
dependencies {
    if (findProject(':script')) {
        implementation project (':script')
    } else {
        debugCompile(name: 'script-debug', ext: 'aar')
        releaseCompile(name: 'script-release', ext: 'aar')
    }
}
```

Makes it easier to build the app together with the framework.

gvr-javascript/app/src/main/java/org/gearvrf/sample/gvrjavascript/GearVRJavascriptActivity.java
```diff
@@ -59,7 +60,7 @@ public void onCreate(Bundle savedInstanceState) {
                 // Attach a script file directly (without using a GVRScriptBundle)
                 GVRScriptFile scriptFile;
                 try {
-                    scriptFile = sm.loadScript(
+                    scriptFile = (GVRScriptFile)sm.loadScript(
```
liaxim: Why the cast? Just use IScriptFile instead of GVRScriptFile.

gvr-javascript/app/build.gradle
xcaostagit: Will do

gvr-javascript/app/src/main/java/org/gearvrf/sample/gvrjavascript/GearVRJavascriptActivity.java
xcaostagit: Will do

PR #643, opened 2018-09-26
Fix gvr-widgetlib-viewer demo 

asg024:
- Fixed the issue with focus handler in model list screen 
- Fixed shader handling imn Model Viewer 
- Removed broken model from the content


ejsamsung:
Updates addressing my comments look good; I'm OK with this getting merged.

PR #646, opened 2018-10-16
gvr-immersivepedia: fix menu rendering for backends that do not support multiview

liaxim:
Monoscopic, daydream and other backends that do not support multiview do not show the menu quads otherwise

Was hard-coded to multiview; using shader template to have the flag set accordingly.


liaxim:
Planning to merge on Friday

PR #653, opened 2018-11-04
Sdc demo

NolaDonato:
This PR is for work on the SDC demo. Currently gvr-avatar is the VR version of the demo and gvr-aravatar is the AR version of the demo.

1. YBot works in both
2. The Cat does not animate (Sahithi)
3. Eva does the drinking animation, does not play other animations (Sahithi)
4. Need to add sample HoloLab models (Nola)
5. need to add GUI to select avatar (Svetlana)


NolaDonato:
replaced by #654

